PR #18, opened 2015-07-20
Fixed local and global aliases

tklink:
The original code to create the alternative name set doesn't work:

    assert_equals = assertEquals

The name `assertEquals` doesn't exist, it's necessary to use `M.assertEquals` instead.

The proposed path creates local aliases first, and then - if required - make them global in the original way.


bluebird75:
Thanks, this was on my todo-list.

PR #21, opened 2015-10-16
more refactoring / code cleanup

n1tehawk:
The first commit in this series is a trivial change to  `assertError()`, the rest deals with factoring out and reusing code.

The goal was to avoid duplication of tasks, leading to smaller - and IMHO more readable - code. It also reduces on the number of string concatenations (especially when constructing error messages).

I've tried to assure that everything still works as expected by making sure that `lua run_unit_tests.lua` and `lua run_functional_tests.lua` are fine (using Lua 5.1.5).

Regards, NiteHawk


bluebird75:
Hi,

Thanks a lot for the contribution. Add unit-tests to your new shiny error formatting functions and I'll merge it happily. Please try to cover all the corner case, the complexity of this functions deserves to be well tested ... especially in a unittest library !

If you feel like you have motivation for further improvements, one next big step is to detect "failures" ( = expected errors) from errors. It's not that difficult, instead of just returning a string with a typical error function when an assertion fails, one needs to return a full object (to be defined) containing the error description (msg, line, file) ... If error still returns a string, it's a true error.

Once that is done, the general framework needs to be improved to report separately errors from failures. The  junit-xml would use this info for example. One difficulty is to properly test that within your own unittest, there is a small problem of failing luaunit execution within a luaunit execution, but nothing difficult to solve I beleive.

n1tehawk:
Hi bluebird75!

`prettystrPadded()` and `error_fmt()` were modelled (and used) in a way that keeps the output of existing tests intact (read: unchanged), so there already was some implicit checking.

I have now added two explicit unit tests to the `TestLuaUnitUtilities` class in _test/test_luaunit.lua_. If you prefer, I can easily merge that commit with the one introducing the two functions; so that this change would become 'atomic'.

Let me know if this is what you had in mind, or if you can think of more/other tests to achieve the desired coverage.

Regards, NiteHawk

n1tehawk:
I have updated the pull request to make all test failures ("expected" errors) explicitly pass through `failure()`. Currently this simply forwards the message to `error()`, but it represents a single function where we might add customizations for the _"failure vs. error"_ distinction.

Regards, NiteHawk

bluebird75:
That's good !

Issue #14, opened 2015-04-23
Update luarocks to use last version


rodfersou:
Sorry, I'm new in Lua world, but luarocks is installing version 2.1, are there a way to update it?

Thanks!

rodfersou:
Other question.. why can't I find version 2.1 at github history and tags?

bluebird75:
Version 2.1 was forked away from my version. I am now in the process of merging everything back together (the fork and my further developments). As soon as this is done, I'll publish it on luarocks.

rodfersou:
@bluebird75 thanks

andreaskoepf:
+1 for making it available on luarocks

n1tehawk:
As of commit c80efe9, an up-to-date LuaRocks version is now available.

bluebird75:
Done after more than one year, see https://luarocks.org/modules/bluebird75/luaunit

PR #22, opened 2015-11-01
Build tool and docs

JJ:
It took me a good while to figure out why nothing in the docs was working, so I finally write this helper Makefile and modified docs a bit. It also refs #14 , which gave me the hint.


bluebird75:
Hi JJ !

Thanks for the contribution. 

However, I already have a central python files responsbile for managing mostly everything ( doit.py ). I prefer you to add an install target to that file rather than a new file and tool to maintain.

It should not be that difficult, your Makefile is super simple. Don't hesitate to ask for support if needed, I promise I will take less than 2 months to answer this time...

Cheers,

Philippe

JJ:
Will try to. Not as good with Python, we oldtimers actually prefer Makefiles :-)

JJ:
Here you go. Will you keep maintaining the Windows version? The build fails.

bluebird75:
Good !

PR #26, opened 2015-12-29
Update and adjust Travis CI integration

n1tehawk:
- update `.travis/` scripts according to https://github.com/moteus/lua-travis-example
- re-enable LuaJIT tests (functional again after 0c0f303)
- add "sudo: false", allowing container-based infrastructure (see https://docs.travis-ci.com/user/workers/container-based-infrastructure/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,29 @@
 language: c
 
+sudo: false
+
 env:
   global:
-    - LUAROCKS=2.2.0-rc1
+    - LUAROCKS=2.2.2
   matrix:
     - LUA=lua5.1
     - LUA=lua5.2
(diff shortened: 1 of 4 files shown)


n1tehawk:
To clarify why `sudo: false` should be added:

It's especially useful for forks of the repo created after 2015-01-01. By default (= with no `sudo` value assigned) these run in "container" mode, causing the current Travis test to always fail. E.g. https://travis-ci.org/n1tehawk/luaunit/branches shows this on `contrib` and `master`. (The main repo has no problem with this, as it existed on Travis long enough to default to "standard" mode.)

bluebird75:
Thanks for luajit support, that was missing indeed !

Issue #25, opened 2015-12-29
Support SetUp/TearDown methods with different names : setup, ...


bluebird75:
Context : LuaUnit version 3 should be backward compatible with LuaUnit v2 which was forked .

LuaUnit v2 supports different names for setup / teardown methods ( see https://github.com/rjpcomputing/luaunit ). LuaUnit v3 should support all the methods :
- names to support : 'setUp' or 'Setup' or 'setup'. 'tearDown' or 'TearDown' or 'teardown'.

Unit-test should be added to verify that this is properly supported.

bluebird75:
task is up for adoption...

n1tehawk:
Hi Philippe!
I'd like you to have a quick glance at 2d657cf. I suggest modifying `LuaUnit.isFunction()` in such a way that it either returns back the function or `nil`. While the result still remains valid as a boolean expression, this additionally would allow us to pick from the supported method names (in order of precedence) using the `or` operator: https://github.com/n1tehawk/luaunit/commit/2d657cf9157ec25afd3885662c4a540145f9d4b5#diff-a7d09041b867f9c9148e0e7a1bd1e7cfR1692

If you approve this, let me know how you'd like to proceed - I can create a 'standalone' pull request for it, or submit it as part of my next PR.

Regards, NiteHawk

n1tehawk:
Btw: Their naming scheme is inconsistent. We need to support `Setup`/`TearDown`, while it should be either `SetUp`/`TearDown` or `Setup`/`Teardown`. The commit I linked had that wrong, I've fixed it in my private branch in the meantime...

bluebird75:
Let's have a wide name coverage to make porting from every other xUnit lua framework easy. Let's support the 4 variants : setup, SetUp, setUp, Setup and teardown, TearDown, tearDown, teardown .

Your commit looks good to me if you add the fourth case, please submit a PR and I'll merge it.

AbigailBuccaneer:
(This issue has been fixed and needs closing.)

PR #27, opened 2016-01-02
Support alternative names for "setUp" and "tearDown" methods

n1tehawk:
This commit addresses #25, and extends the appropriate unit test.

Note that we test for function names in the same order (= precedence)
that the forked LuaUnit v2.1 used. We also support the additional
names "SetUp" and "Teardown" for sake of consistency.


n1tehawk:
(You might want to apply something like 4835eed first, to have the build checks pass again.)

PR #31, opened 2016-01-06
Fix _is_table_equal with table keys

### LINKED ISSUE #30: `assertEquals` can't match tables with table keys
```Lua
> luaunit.assertEquals({ [{}] = {} }, { [{}] = {} })
```
```
stdin:1: expected: {{}={}}
actual: {{}={}}
stack traceback:
	[C]: in function 'error'
	./luaunit.lua:536: in function 'failure'
	./luaunit.lua:598: in function 'assertEquals'
	stdin:1: in main chunk
	[C]: in ?
```

AbigailBuccaneer:
See new unit tests for cases this fixes.

Fixes #30.

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -22,6 +22,7 @@ assertEquals( expected, actual ).
 ]]--
 M.ORDER_ACTUAL_EXPECTED = true
 M.PRINT_TABLE_REF_IN_ERROR_MSG = false
+M.TABLE_EQUALS_KEYBYCONTENT = true
 M.LINE_LENGTH=80
 
 -- set this to false to debug luaunit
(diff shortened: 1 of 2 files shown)


n1tehawk:
Table "equality" always is a somewhat delicate question in Lua, as tables are passed by reference. Therefore it's well possible that two tables `a` and `b` are not considered identical (i.e. Lua will report `a ~= b`), despite them both "holding" the same content.

This patch goes quite some way to try and match up table-type keys (by their content) when comparing tables with `_is_table_equal()`. While that may be useful under special circumstances, I'm not convinced that I'd consider this "natural" for a *key* value comparison, or would even want it as default behavior.

Do you think this could be made optional / configurable via some `M.OPTION` setting?

Regards, NiteHawk

PR #36, opened 2016-01-10
Distinguish "errors" from "failures"

n1tehawk:
LuaUnit was a bit sloppy when it came to detecting "expected" (test) failures from runtime errors, either within the framework itself or originating from user code. These patches will hopefully solve most of the issues involved.


bluebird75:
Thanks again

bluebird75:
Next step is to introduce notion of skipping test. That's slightly more difficult !

Issue #29, opened 2016-01-06
`prettystr` fails on tables with table/boolean keys


AbigailBuccaneer:
```Lua
> local prettystr = (require "luaunit").prettystr
> prettystr{ [true] = true } -- okay
> prettystr{ [{}] = {} } -- okay
> prettystr{ string = "string", [true] = true } -- okay
> prettystr{ string = "string", [{}] = {} } -- okay
> prettystr{ string = "string", 1 } -- okay
> prettystr{ 1, [true] = true } -- error: attempt to compare boolean with number
> prettystr{ 1, [{}] = {} } -- error: attempt to compare table with number
> prettystr{ [true] = true, [{}] = {} } -- error: attempt to compare boolean with table
```

It looks like if the set of the types of keys contains more than one of `{ number, boolean, table }` then `table.sort` throws an error.

`luaunit.private.__genSortedIndex` splits the table into two parts - those with string keys, and those with all other keys - and then sorts those two parts separately. `table.sort` can't handle sorting on heterogeneous keys.

The stack trace (for all of these errors):
```
stack traceback:
	[C]: in function 'sort'
	./luaunit.lua:73: in function '__genSortedIndex'
	./luaunit.lua:105: in function 'for iterator'
	./luaunit.lua:353: in function <./luaunit.lua:333>
	(...tail calls...)
	./luaunit.lua:404: in function 'prettystr'
	stdin:1: in main chunk
	[C]: in ?
```

bluebird75:
Closed, with the solution of nitehawk

PR #38, opened 2016-01-10
Fix genSortedIndex for heterogeneous keys

n1tehawk:
This pull request has the suggested changes that are still missing from closing #29, including the new comparison/sort function for arbitrary values (`crossTypeSort`) that the modified `__genSortedIndex` makes use of.

It also adds a commit to exploit the new features for improvements in `sortedNext()` and `_table_tostring()`.

(Note: merging this PR will obsolete #32.)


bluebird75:
Very nice as usual

PR #42, opened 2016-01-31
Some small fixes and a new "--error" option

n1tehawk:


luaunit.lua
```diff
@@ -1759,6 +1764,12 @@ local LuaUnit_MT = { __index = M.LuaUnit }
         node.duration = os.clock() - node.startTime
         node.startTime = nil
         self.output:endTest( node )
+        -- "--error" option: runtime errors will lead to program termination
+        if self.quitOnError and node:isError() then
+            print("\nERROR during LuaUnit test execution:\n" .. node.msg)
+            print("LuaUnit ABORTING (as requested by --error option)")
+            os.exit(-2)
```
bluebird75: I would prefer that the exit is done after calling endClass() and endSuite(). Today, it makes no difference but class-level setup/teardown and suite-level setup/teardown are on the TODO list.

Please use a flag to trigger an early exit but let the suite close correctly. Displaying the global report might also be a good idea, you may want to know what has happened before.

n1tehawk: That's a valid concern, and I have modified the commit accordingly. As of now, running LuaUnit:endSuite() also causes the outputter to display the results. However with test execution aborted, the assumption `passedCount = testCount - notPassedCount` may no longer hold; so we need a slightly more sophisticated test count management.

Regards, NiteHawk

Issue #44, opened 2016-03-22
Preparing v3.2 release


bluebird75:
I am preparing a v3.2 release for soon. 
@AbigailBuccaneer  @n1tehawk if you have some pending modifications which you would like to push, now is the good time !

Cheers,

Philippe

n1tehawk:
Hi Philippe!

I have nothing urgent on my agenda, but would like to propose two small patches. I'll create a PR for those later, after I have caught up on your latest changes (and rebased my stuff accordingly).

(Nevertheless, I do not consider either of these patches necessary or vital for the upcoming release.)

Regards, NiteHawk

n1tehawk:
I have added two more commits: first adds a test case for `--failure`, the second addresses the TODO related to grammatical number in `statusLine()`. Feel free to cherry-pick whatever you consider useful.

Regards, NiteHawk

n1tehawk:
I'm not sure if you still are collecting / reviewing changes for the upcoming release. If so, I have two more optional patches in my [contrib branch](https://github.com/n1tehawk/luaunit/commits/contrib) that you might consider for inclusion. (Let me know if you want me to create a PR for them.)

Regards, NiteHawk

n1tehawk:
I think that with adding static analysis from #55 (and possibly the fixes in #56) a point will be reached where it would make sense to tag v3.2.

bluebird75:
Sorry for the delay, real life has kept me very busy over the last few months.

The only thing really holding me back for v3.2 release is that documentaiton has not been updated to reflect the latest config for comparing recursive tables and for the --error flag.

bluebird75:
Release done.

PR #45, opened 2016-03-22
My changes for v3.2 release

n1tehawk:
Related to #44 - These are the patches I'd like you to consider for merging.

Regards, NiteHawk

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,7 @@ test/unitTests*.xml
 test/unitTests*.txt
 test/legacyExample.txt
 test/legacyExampleError.txt
+test/legacyExampleFailure.txt
 test/compat_luaunit_v2x.txt
 test/errFailPass*.txt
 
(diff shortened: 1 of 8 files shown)


bluebird75:
Nice for the addStatus, I also had that in mind at some point.

For the io.write vs io.stdout:write, would you care to elaborate on the difference for a Lua newbie like me ?

n1tehawk:
> For the io.write vs io.stdout:write, would you care to elaborate on the difference for a Lua newbie like me ?

There is none - except that `io.write()` is shorter. See e.g. http://www.lua.org/manual/5.1/manual.html#pdf-io.write

bluebird75:
I prefer that explicit form then. That's my python background where "explicit is better than implicit" is a key value....

n1tehawk:
Okay. To straighten this out now, I'll close this PR and create another one - with 2016fc6 and 3cd17f0 rebased on top of current master.

Regards, NiteHawk

PR #46, opened 2016-03-26
Some code cleanup

n1tehawk:
~~Two~~ more (smaller) commits based on current master

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -1862,22 +1862,6 @@ local LuaUnit_MT = { __index = M.LuaUnit }
         error( 'No such format: '..outputType,2)
     end
 
-    function M.LuaUnit:setVerbosity( verbosity )
-        self.verbosity = verbosity
-    end
-
-    function M.LuaUnit:setFname( fname )
-        self.fname = fname
-    end
-
-    function M.LuaUnit:setQuitOnError( value )
-        self.quitOnError = value
-    end
-
-    function M.LuaUnit:setQuitOnFailure( value )
-        self.quitOnFailure = value
-    end
-
     --------------[[ Runner ]]-----------------
 
     function M.LuaUnit:protectedCall(classInstance, methodInstance, prettyFuncName)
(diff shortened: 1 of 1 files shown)


luaunit.lua
```diff
@@ -1279,19 +1264,12 @@ then OK or FAILED (failures=1, error=1)
...
 
     function TextOutput:new()
-        local t = {}
-        t.runner = nil
-        t.result = nil
-        t.errorList ={}
-        t.verbosity = M.VERBOSITY_DEFAULT
-        setmetatable( t, TextOutput_MT )
-        return t
+        return setmetatable(
```
bluebird75: Something gets lost along the way : the explicit (useless) initialization was a good way to remind the reader of what fields are actually used within the pseudo-object. With the new syntax, it's not clear that fields runner and result are actually available....

Applies also to above classes.

n1tehawk: There same goes for other private fields that get initialized from within `LuaUnit.startSuite()` (https://github.com/bluebird75/luaunit/blob/b5fd780ec59be20c3664809094f8b4d1583549bc/luaunit.lua#L1735-L1738), namely `verbosity` and `fname`.

Setting them "across" classes like that is probably a bad idea anyway. Maybe we should include those options explicitly in the constructor - something like `self.output = self.outputType:new(self, self.result, self.verbosity, self.fname)`.

Unfortunately, implementing it like that currently opens just another can of worms. There are unit tests that rely on the side effects (assigning the outputter's fields) of `LuaUnit.startSuite()` - using a custom outputter like in https://github.com/bluebird75/luaunit/blob/b5fd780ec59be20c3664809094f8b4d1583549bc/test/test_luaunit.lua#L2475-L2476 therefore tends to break easily...

n1tehawk:
_Post scriptum:_ Some of the concerns expressed above should now be solved via 6a18b42 and b531c64. The "base" class constructor for outputters (`genericOutput.new(()`) now takes care of copying fields from the parent class ("runner"), and since the test mockup classes inherit from it, the behavior is consistent. `LuaUnit.startSuite()` no longer assigns any ouputter fields directly.

PR #53, opened 2016-06-04
Add 'local' to variable declarations

AlfonZ42:
Hello.

I've added some _local_ modifiers for variable declarations.

Also, what do think about enabling _strict_ module for testing? See https://github.com/AlfonZ42/luaunit/compare/locals...AlfonZ42:strict

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -1986,8 +1986,8 @@ end
                     error( 'Instance must be a table or a function, not a '..type(instance)..', value '..prettystr(instance))
                 end
                 if M.LuaUnit.isClassMethod( name ) then
-                    className, methodName = M.LuaUnit.splitClassMethod( name )
-                    methodInstance = instance[methodName]
+                    local className, methodName = M.LuaUnit.splitClassMethod( name )
+                    local methodInstance = instance[methodName]
                     if methodInstance == nil then
                         error( "Could not find method in class "..tostring(className).." for method "..tostring(methodName) )
                     end
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -1976,9 +1976,10 @@ end
         -- expand all classes (provided as {className, classInstance}) to a list of {className.methodName, classInstance}
         -- functions and methods remain untouched
         local result = {}
+        local name, instance, className, methodName, methodInstance
 
         for i,v in ipairs( listOfNameAndInst ) do
-            local name, instance = v[1], v[2]
+            name, instance = v[1], v[2]
             if M.LuaUnit.asFunction(instance) then
                 table.insert( result, { name, instance } )
             else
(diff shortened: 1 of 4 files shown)


luaunit.lua
```diff
@@ -1986,8 +1986,8 @@ end
                     error( 'Instance must be a table or a function, not a '..type(instance)..', value '..prettystr(instance))
                 end
                 if M.LuaUnit.isClassMethod( name ) then
-                    className, methodName = M.LuaUnit.splitClassMethod( name )
-                    methodInstance = instance[methodName]
+                    local className, methodName = M.LuaUnit.splitClassMethod( name )
+                    local methodInstance = instance[methodName]
```
n1tehawk: For improved performance, it's usually advisable not to declare local variables within loops. Actually this applies to `name` and `instance` here, too. Ideally all four should be declared (`local name, instance, className, methodName`) _above_ the for-loop.

AlfonZ42:
Ok, I've moved almost all local variable declarations out of loops.
Where is the threshold between having them out of loop for better performance and having them inside the loop for better clarity?

n1tehawk:
I've also come across these problems and did something similar in https://github.com/n1tehawk/luaunit/commit/0d440287c33faa75b313cfbd3ed1e0a302e32c17.

>Also, what do think about enabling _strict_ module for testing? See https://github.com/AlfonZ42/luaunit/compare/locals...AlfonZ42:strict

I agree that we should probably take greater care to avoid missing `local` (possibly introducing unwanted globals), and might make use of the CI framework to ensure this policy.

Using the "strict" module is one possible approach, another might be static analysis of the code - e.g. `luac -p -l luaunit.lua | grep SETGLOBAL`. See http://lua-users.org/wiki/DetectingUndefinedVariables Approach No. 2.

Regards, NiteHawk

n1tehawk:
>Where is the threshold between having them out of loop for better performance and having them inside the loop for better clarity?

It's a question of scope, and more a personal preference instead of a rule set in stone.

Actually, I couldn't quite remember it where I once found that tip - so I tried to look it up on the Lua user's wiki, without success... I even came across http://lua-users.org/lists/lua-l/2011-09/threads.html#00521, which indicates that local with loops **doesn't matter at all** - and suggests my "best practice" isn't particularly useful at all. So apologies for causing a little stir here, and for any inconvenience this might have caused. :blush:

bluebird75:
I am relieved to see that most changes are inside test files, which are less sensitive.

Still, any automation in this area would definitely be welcomed. I am not familiar with "strict" can you point me to any references ?

n1tehawk:
Philippe, the `strict` module is described in section "_Approach No. 1: Run-time Checking_" of above link to the Lua user's wiki.

I've been experimenting a bit with this, and have come to the conclusion that I like the "static analysis" approach better:

1. It's less intrusive (requires no modification of existing sources), and mostly independent of the specific Lua(JIT) version in use.

2. Since it doesn't rely on runtime detection, It can find problematic assignments even if they aren't executed. This may apply to code paths that are optional, or rarely visited (e.g. error handlers).

I'm currently preparing an addon that does some simple checking of the Lua bytecode (inspired by the second approach described in the wiki), and can be run as an extra `run_static_tests.lua` build step. Once it's ready, I'll open another PR for that.

Regards, NiteHawk

Issue #48, opened 2016-04-20
luaunit.assertEquals() causes stack overflow when passing self-referencing table


KosnIre:
When writing a test that should verify a table, I used `luaunit.assertEquals(table1, table2)`

A stack overflow occurs because `table1` contains a key with a value set to `table1`. The issue is a result of logic in the `_is_table_equals()` function in luaunit.lua (currently line 510).

While debugging, I have added a print statement just above this recursive call at line 527:

    print("_is_table_equals Testing:", actual, k, v)
    if not _is_table_equals(v, expected[k]) then
        return false
    end

resulting in the following log output:

_is_table_equals Testing:       table: 0AA87F28 removeEventListener     function: 0A4F5070
_is_table_equals Testing:       table: 0AA87F28 addEventListener        function: 0A4F52F0
_is_table_equals Testing:       table: **0AA87F28** __index table: **0AA87F28**
_is_table_equals Testing:       table: 0AA87F28 removeEventListener     function: 0A4F5070
_is_table_equals Testing:       table: 0AA87F28 addEventListener        function: 0A4F52F0
_is_table_equals Testing:       table: 0AA87F28 __index table: 0AA87F28
_is_table_equals Testing:       table: 0AA87F28 removeEventListener     function: 0A4F5070
_is_table_equals Testing:       table: 0AA87F28 addEventListener        function: 0A4F52F0
_is_table_equals Testing:       table: 0AA87F28 __index table: 0AA87F28
_is_table_equals Testing:       table: 0AA87F28 removeEventListener     function: 0A4F5070
_is_table_equals Testing:       table: 0AA87F28 addEventListener        function: 0A4F52F0
_is_table_equals Testing:       table: 0AA87F28 __index table: 0AA87F28
_is_table_equals Testing:       table: 0AA87F28 removeEventListener     function: 0A4F5070
_is_table_equals Testing:       table: 0AA87F28 addEventListener        function: 0A4F52F0
_is_table_equals Testing:       table: 0AA87F28 __index table: 0AA87F28
_is_table_equals Testing:       table: 0AA87F28 removeEventListener     function: 0A4F5070
_is_table_equals Testing:       table: 0AA87F28 addEventListener        function: 0A4F52F0
_is_table_equals Testing:       table: 0AA87F28 __index table: 0AA87F28
_is_table_equals Testing:       table: 0AA87F28 removeEventListener     function: 0A4F5070
[...]

PR #57, opened 2016-06-09
Improved _table_tostring and _is_table_equals functions for some recursive tables

AlfonZ42:
Hello.

I've made some fixes and changes when dealing with some recursive tables. In a sense it's a follow-up to #48 . The goal was to make a deep-copied recursive table equal to the original one.

Changes the output of

    prettystr( {["foo\nbar"] = 1} )

from

    {"foo
    bar"=1}

to

    {"foo\nbar"=1}

Changes the output of

    E = {}; A = {"A", E}; B = {A}; A[B] = 1; prettystr(A)

from (if _PRINT_TABLE_REF_IN_ERROR_MSG=false_)

    {"A", {}, <table: addrB> {<table: addrA> {"A", <table: addrE> {}, <table: addrB>=1}}=1}

and from (if _PRINT_TABLE_REF_IN_ERROR_MSG=true_)

    {"A", <table: addrE> {}, <table: addrB> {<table: addrA> {"A", <table: addrE> {}, <table: addrB>=1}}=1}

to

    <table: addrA> {"A", <table: addrE> {}, <table: addrB> {<table: addrA>}=1}

(actual addresses replaced with placeholders)
This particular test case uses a hack that temporarily overwrites LINE_LENGTH - 
addresses printed out by _tostring(table)_ are longer with 64bit Lua, which triggers multi-line display of the string. With 32bit Lua the string is shorter, so it stays in one line. The threshold in this example seems to be 4 addresses (3 addresses can fit into one line, 4 addresses will be too long).
This seems to be the first test case that is affected by this, so I'm not sure if it's the correct way to deal with this.

Changes the result of test case

    A={}; B={}; A.self = A; B.self = B; assertNotEquals(A, B)

from passing to failing, i.e.

    assertEquals(A, B)

is now passing.

Changes the result of test case

    A,B,C={},{},{}; A.ref=C; B.ref=A; C.ref=B; assertNotEquals(A, B)

from passing to failing, i.e.

    assertEquals(A, B)

is now passing.

The *_is_table_equals* function now tracks recursion of both _actual_ and _expected_ tables and their pairings. If this particular pairing is detected later, it's declared equal, since all other fields up to this time were equal as well (otherwise the inequality will stop the descent and return back).
The descent ends if either _actual_ or _expected_ were previously encountered (and are thus in the appropriate _recursions_ table). If only one of them is recurring, they are declared not equal, even though they could be equal (contents-wise).
Thus there is still room for improvement, although I am not sure how likely is that such convoluted case will be encountered.

Also, code-formatting-style-wise, there are some instances of

    if condition then statement end

that might be split into separate lines, so line code coverage (that @n1tehawk has in one of their branches) could report them in more details.

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -551,15 +551,18 @@ local function _is_table_equals(actual, expected, recursions)
                 if not candidates then return false end
                 for i, candidate in pairs(candidates) do
                     if _is_table_equals(candidate, k) then
-                        found = candidate
-                        -- Remove the candidate we matched against from the list
-                        -- of candidates, so each key in actual can only match
-                        -- one key in expected.
-                        candidates[i] = nil
-                        break
+                        if _is_table_equals(actual[candidate], v) then
+                            found = candidate
+                            -- Remove the candidate we matched against from the list
+                            -- of candidates, so each key in actual can only match
+                            -- one key in expected.
+                            candidates[i] = nil
+                            break
+                        end
+                        -- keys match but values don't, keep searching
                     end
                 end
-                if not(found and _is_table_equals(actual[found], v)) then
+                if not found then
                     -- Either no matching key, or a different value
                     return false
                 end
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -414,10 +414,15 @@ local function _table_tostring( tbl, indentLevel, printTableRefs, recursionTable
             -- for the sequential part of tables, we'll skip the "<key>=" output
             entry = ''
             seq_index = seq_index + 1
+        elseif recursionTable[k] then
+            -- recursion in the key detected
+            recursionTable.recursionDetected = true
+            entry = "<"..tostring(k)..">="
         else
             entry = _table_keytostring( k ) .. "="
         end
-        if recursionTable[v] then -- recursion detected!
+        if recursionTable[v] then
+            -- recursion in the value detected!
             recursionTable.recursionDetected = true
             entry = entry .. "<"..tostring(v)..">"
         else
@@ -512,6 +517,14 @@ local function _is_table_items_equals(actual, expected )
     return true
 end
 
+local function _table_copy(tbl)
+    local newtbl = {}
+    for k, v in pairs(tbl) do
+        newtbl[k] = v
+    end
+    return newtbl
+end
+
 local function _is_table_equals(actual, expected, recursions)
     local type_a, type_e = type(actual), type(expected)
     recursions = recursions or {}
@@ -550,13 +563,19 @@ local function _is_table_equals(actual, expected, recursions)
                 local candidates, found = actualTableKeys[#k], nil
                 if not candidates then return false end
                 for i, candidate in pairs(candidates) do
-                    if _is_table_equals(candidate, k) then
-                        if _is_table_equals(actual[candidate], v) then
+                    -- use temporary recursions table for testing, revert to saved version if matching failed
+                    local recursions_temp = _table_copy(recursions)
+                    if _is_table_equals(candidate, k, recursions_temp) then
+                        if _is_table_equals(actual[candidate], v, recursions_temp) then
                             found = candidate
                             -- Remove the candidate we matched against from the list
                             -- of candidates, so each key in actual can only match
                             -- one key in expected.
                             candidates[i] = nil
+                            -- confirm new recursions table
+                            -- note: new table will be used for deeper levels, will not propagate back
+                            --   haven't found counter-example yet where this causes problems
+                            recursions = recursions_temp
                             break
                         end
                         -- keys match but values don't, keep searching
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -543,7 +543,28 @@ local function _is_table_equals(actual, expected, recursions_a, recursions_e)
     recursions_a = recursions_a or {}
     recursions_e = recursions_e or {}
 
-    if (type_a == 'table') and (type_e == 'table') and not recursions_a[actual] and not recursions_e[expected] then
+    if type_a == type_e and actual == expected then
+        return true
+
+    elseif recursions_a[actual] ~= nil and recursions_a[actual][expected] then
+        -- the pair was already visited, consider them equal
+        --  that is, the pair has already been confirmed equal (and we return the same)
+        --  or we are in the process of deciding it and found a loop (return true so the process can continue)
+        return true
+
+    elseif recursions_e[expected] ~= nil and recursions_e[expected][actual] then
+        -- should not get here, recursions tables should be symmetric
+        print("\n\nactual=" .. prettystr(actual))
+        print("expected=" .. prettystr(expected))
+        print("recursions_a=" .. prettystr(recursions_a))
+        print("recursions_e=" .. prettystr(recursions_e))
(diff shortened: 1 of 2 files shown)


AlfonZ42:
Changed the recursion detection condition to check if the pair *(actual,expected)* has been visited, instead of whether either of them has been visited. This should make comparing the tables in the fullest way possible.
Since *actual* and *expected* tables are finite, the number of sub-tables they consist of is finite as well, thus the number of pairs of *(actual,expected)* is finite as well, and the recursion will stop eventually.

This changes the output of some test cases (that were previously considering asymetrically looping tables to be not equal).

n1tehawk:
Hi AlfonZ42!

I think that last idea of testing the `_is_table_equals()` recursions based on the specific _(actual,expected)_ pairs is really great, since it gives a very solid stop criterion.

I understand that your code has grown "organically" before arriving at that point - but I'd prefer to implement that concept in a more fundamental way. For this I have reworked the _recursions_ table by adding a metatable with two "methods" that help a lot with the recursion / result caching logic: see https://github.com/n1tehawk/luaunit/commit/db3a691bf06e1dfb4c0a3810fd08653b5eca3065. This keeps the actual `_is_table_equals()` code much cleaner.

---
>Also, code-formatting-style-wise, there are some instances of
>
>`if condition then statement end`
>
>that might be split into separate lines, so line code coverage (that @n1tehawk has in one of their branches) could report them in more details.

That's a useful suggestion, thanks. I did not perceive any real issues with it so far, but will definitely keep an eye on that.

I've been going through your commits over the last days, and will comment on certain aspects individually - please bear with me...

Regards, NiteHawk

n1tehawk:
My personal view is that commit [0f55470](https://github.com/bluebird75/luaunit/pull/57/commits/0f55470e6070fdede84c4b4a7f7bbaf42bc6a3ee) shows how things turn ugly quickly with "table equality" matching of table-type keys: multiple (different) instances of tables holding identical content, and then pointing to different values...

Nevertheless, the proposed fix / commit in itself is correct of course. Originally I was concerned that the repeating "candidate" matching loops would mess up the `recursions` (tracking) table quickly, but the _(actual,expected)_ pairs approach seems to solve that quite nicely.

---
I'd prefer it if [e85e6a6](https://github.com/bluebird75/luaunit/pull/57/commits/e85e6a64fdcef7428d69e0d6a8614d2a729cb2d8) was split in half - one commit for the `_table_tostring()` fix, plus the more strict `test_prettystrTableRecursion()`; and another one for the `_is_table_equals()` recursions fix (and test cases).

BTW: While we're at it, I think we should extend the patterns in _test_prettystrTableRecursion()_ to include the `<table` part and the closing `>` so the repeating captures (`%1`, `%2`, ...) are shorter.

I've done this in https://github.com/n1tehawk/luaunit/commit/b6858066ee5af3c7ad27352bf147c7ccb22cdf4b and https://github.com/n1tehawk/luaunit/commit/abaf59469c4ad3dc32b014871c786d74f2065faa.

n1tehawk:
@[c56f877](https://github.com/bluebird75/luaunit/pull/57/commits/c56f877d649b26316b337be25de204ec9f3a44f9)
>This particular test case uses a hack that temporarily overwrites LINE_LENGTH - addresses printed out by _tostring(table)_ are longer with 64bit Lua, which triggers multi-line display of the string. With 32bit Lua the string is shorter, so it stays in one line. The threshold in this example seems to be 4 addresses (3 addresses can fit into one line, 4 addresses will be too long).
This seems to be the first test case that is affected by this, so I'm not sure if it's the correct way to deal with this.

I think the actual root cause for this problem sits at a deeper level: `prettystring(t, true)` simply won't respect the requested _"keeponeline"_ flag for complex tables (exceeding the line length).

In other words: **If** we make `_table_tostring` (and thus `prettystr`) obey that flag properly first, the solution to this problem gets trivial: just use `lu.prettystr(t9, true)`. See https://github.com/n1tehawk/luaunit/commit/cc010de4751ccd0623e9bb8efd6df7f640237ca4 and https://github.com/n1tehawk/luaunit/commit/fcee4f3280279571f0c9cffa6ffbbc0738e2ba16.

n1tehawk:
The remaining commits of your series is where I feel somewhat at a loss. Personally I dislike stretching "table equality" to extend to "loops in the same way", and find this a concept that's rather hard to grasp. I also have to admit that I get lost in your adjustments to the matching code in `_is_table_equals`, especially the introduction of two separate recursion tables for _actual_ and _expected_.

From my (limited?) point of view the one key change you make affects the "default" value `true` assumed for matching two tables: https://github.com/bluebird75/luaunit/pull/57/commits/fd16aa772677ed5a8d663ea37bee10a252b7e9a6#diff-a7d09041b867f9c9148e0e7a1bd1e7cfR553.

I have tried to achieve something similar with https://github.com/n1tehawk/luaunit/commit/508351f3cbf7ea27c550680ad3f1b3f69d136a1a. After setting `M.TABLE_EQUALS_RECURSIONDEFAULT = true` all your newly introduced unit tests seem to pass - see https://github.com/n1tehawk/luaunit/commit/81b5c5a2252445cf8f448fa52c09860f73b65567. But again: I'm still not convinced that this is "the right way" to match tables, and wouldn't want that enabled by default. (Maybe because I also simply fail to see how my code would benefit from these cases.)

Regards, NiteHawk

AlfonZ42:
I think this example might show how I view table equality with loops:

	local A, B, C, D = {1}, {1}, {2}, {2}
	lu.assertEquals(A, B)
	lu.assertEquals(C, D)
	A.ref = C
	B.ref = D
	lu.assertEquals(A, B)
	A.self = A
	B.self = B
	lu.assertEquals(A, B) -- fails with lu.TABLE_EQUALS_RECURSIONDEFAULT=false

That is, setting field _ref_ of tables _A_ and _B_ to non-identical but equal tables _C_ and _D_ doesn't change the equality of _A_,_B_. Similarly, one could expect the same if _C_=_A_ and _D_=_B_.

The two tracking tables were previously needed when the stop criterion was _either table is recurring_ (since it was possible to create two tables such that _assertEquals(A,B)_ and _assertNotEquals(B,A)_).
Similarly for temporary tracking tables - without them, the negative result of comparing tables affected the result.
Now that the criterion changed to _(actual,expected)_ pair, they are no longer needed, and the new _recursions_ table holds all the information needed to evaluate that.

n1tehawk:
I have added a [20160724_refactor-pull-request-57](https://github.com/n1tehawk/luaunit/commits/20160724_refactor-pull-request-57) branch to my repo, that has the previous changes (*20160612_refactor-pull-request-57*) rebased onto current *master*. Might help with reviewing...

bluebird75:
Functionality has been merged as part of #65 . Feel free to comment and repropose if you think more improvements are needed.

n1tehawk:
After merging #65, as far as I can tell only a change to the default value for matching tables recursively remains - this was mentioned before (see above), and would allow the test cases that were originally proposed to pass. I've rebased the two related commits onto current master and kept them in [20160812_refactor-pull-request-57](https://github.com/n1tehawk/luaunit/commits/20160812_refactor-pull-request-57) for reference.

PR #64, opened 2016-07-25
LuaCov coverage tests support

n1tehawk:
This adds some (very) basic support for running coverage tests on
LuaUnit, using LuaCov (https://github.com/keplerproject/luacov).
LuaCov can be installed using LuaRocks (luarocks install luacov).

Running tests with LuaCov involves loading the "luacov" module
first. run_unit_tests.lua requires no modification and can simply
be run as "$LUA -lluacov run_unit_tests.lua".

However, the functional tests in run_functional_tests.lua do their
own invocation of the Lua interpreter, which needs to be adjusted.
To achieve the desired result, the script now accepts a new option
"--coverage" which will modify the Lua calls accordingly. So to get
coverage analysis, use "./run_functional_tests.lua --coverage".

LuaUnit execution itself should not be affected by this in any way,
except for data collection (in luacov.stats.out). The coverage data
normally needs post-processing to give useful results, see LuaCov
usage instructions for details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,9 @@ obj/
 [Rr]elease*/
 _ReSharper*/
 
+# LuaCov output (stats and report)
+luacov.*.out
+
 # LuaUnit ignores
 [Tt]est[Rr]esult*
 *.sublime-project
--- a/run_functional_tests.lua
+++ b/run_functional_tests.lua
@@ -666,6 +666,10 @@ end
 
 
 local function main()
+    if arg[1] == '--coverage' then
+        LUA = LUA .." -lluacov" -- run tests with LuaCov active
+        table.remove(arg, 1)
+    end
     if arg[1] == '--update' then
         if #arg == 1 then
             -- generate all files


bluebird75:
Nice. Looking forward for the next step with the covering badge...

n1tehawk:
>Nice. Looking forward for the next step with the covering badge...

If you're eager to have that shiny badge on the front page, that should now be simple enough... I'm using the https://coveralls.io/ service, so the first step would be to sign up there (can be done with your Github account) and grant access to the _LuaUnit_ repo ("add repo").

You'll also need a "reporter" to upload coverage data to _coveralls.io_. Luckily [luacov-coveralls](https://github.com/moteus/luacov-coveralls) is readily available (via LuaRocks) and simple to use. I have limited the analysis to `luaunit.lua`, which is the only file we're really interested in. To complete this step, you could cherry-pick the changes to `.travis.yml` from https://github.com/n1tehawk/luaunit/commit/dac8fbad4f62ee1866330469507b7878563b5dbe.

Commit the result, and check if _coveralls.io_ starts to report coverage. If it does, you can use the badge markdown they provide to update `README.md`. Voilà!

n1tehawk:
Well done. :+1: I noticed that you also pulled in the separate commit that I had used to push the coverage figures up another notch...

bluebird75:
Yes. I don't know how you manage to reach 100% though. There is still one if that I can replace with an assert, but the invalid state verification can not be removed...

n1tehawk:
I'm not sure why it doesn't "round up" and show 100% on the badge for this repo. (I had expected that the badge would have updated to 100% by now after adding the above commit.)

I'm not at a 100% percentile either - still missing the exact same two lines as here (i.e. 1030/1032 lines covered = 99,806%). IIRC, for a while I had 99% displayed too - maybe it takes a few commits/builds for coveralls to accept that as 100%. My impression is the "rounding up" should take place as long as we keep the ration at or above 99.5%.

PR #65, opened 2016-07-25
Improve _is_table_equals() function and fix various errors

n1tehawk:
This pull request incorporates a significant portion of the changes proposed in #57. Merging this would fix a number of actual errors (mostly related to table recursion), and should make it easier to examine - and possibly find a consensus on - the remaining changes.

@AlfonZ42 I'd especially like your feedback on this set of patches. Do you consider these useful in moving forward with your pull request?

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -551,13 +551,10 @@ local function _is_table_equals(actual, expected, recursions)
         for k, v in pairs(actual) do
             if M.TABLE_EQUALS_KEYBYCONTENT and type(k) == "table" then
                 -- If the keys are tables, things get a bit tricky here as we
-                -- can have _is_table_equals(k1, k2) and t[k1] ~= t[k2]. So we
-                -- collect actual's table keys, group them by length for
-                -- performance, and then for each table key in expected we look
-                -- it up in actualTableKeys.
-                local count = #k
-                if not actualTableKeys[count] then actualTableKeys[count] = {} end
-                table.insert(actualTableKeys[count], k)
+                -- can have _is_table_equals(t[k1], t[k2]) despite k1 ~= k2. So
+                -- we first collect table keys from "actual", and then later try
+                -- to match each table key from "expected" to actualTableKeys.
+                table.insert(actualTableKeys, k)
             else
                 if not _is_table_equals(v, expected[k], recursions) then
                     return false -- Mismatch on value, tables can't be equal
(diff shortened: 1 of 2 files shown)


n1tehawk:
P.S.: I disliked the fact that the last patch was only adding test cases and most of its original commit message had become meaningless (as some issues were already addressed by previous changes). I've now squashed the two additional tests into the commit that actually fixes the recursion problem (1894f07).

luaunit.lua
```diff
@@ -529,23 +529,73 @@ local function _is_table_items_equals(actual, expected )
...
+        -- Store cached value for a specific (actual,expected) pair.
+        -- Returns the value, so it's easy to use for a "tailcall" (return ...).
+        store = function(t, actual, expected, value, asymmetric)
+            local subtable = t[actual]
+            if not subtable then
+                subtable = {}
+                t[actual] = subtable
+            end
+            subtable[expected] = value
+
```
bluebird75: Is there any reason why we can't use {actual,expected} as key of one table ? The subtable approach looks less readable to me....

test/test_luaunit.lua
```diff
@@ -690,9 +690,13 @@ TestLuaUnitAssertions = { __class__ = 'TestLuaUnitAssertions' }
         lu.TABLE_EQUALS_KEYBYCONTENT = false
         assertFailure( lu.assertEquals, {[{}] = 1}, { [{}] = 1})
         assertFailure( lu.assertEquals, {[{one=1, two=2}] = 1}, { [{two=2, one=1}] = 1})
+        assertFailure( lu.assertEquals, {[{1}]=2, [{1}]=3}, {[{1}]=3, [{1}]=2} )
         lu.TABLE_EQUALS_KEYBYCONTENT = true
         lu.assertEquals( {[{}] = 1}, { [{}] = 1})
         lu.assertEquals( {[{one=1, two=2}] = 1}, { [{two=2, one=1}] = 1})
+        lu.assertEquals( {[{1}]=2, [{1}]=3}, {[{1}]=3, [{1}]=2} )
+        -- try the other order as well, in case pairs() returns items reversed in the test above
+        lu.assertEquals( {[{1}]=2, [{1}]=3}, {[{1}]=2, [{1}]=3} )
```
bluebird75: This gets really braindead no ? What sense does it have to have inside a table two keys with the same table content. I am glad that LuaUnit can deal sanely with this but using table as keys is really ... not sane ! :-)

bluebird75:
Patch looks good and non controversial to me. @AlfonZ42 looking forward to your feedback...

luaunit.lua
n1tehawk: Lua itself has no support (language/syntax construct) for multi-dimensional tables, so some sort of workaround is required - see e.g. [Matrices and Multi-Dimensional Arrays](http://www.lua.org/pil/11.2.html).

test/test_luaunit.lua
n1tehawk: Yes, as I commented earlier: things tend to turn ugly quickly with "table equality" matching of table-type keys.

But I think @AlfonZ42 has some valid point here: If we accept the concept of "equality" based on table content (`M.TABLE_EQUALS_KEYBYCONTENT = true`, dating back to #30), then the example in [his comment](https://github.com/bluebird75/luaunit/pull/57#issuecomment-226185977) makes sense. When supporting this, we probably need to go all the way...

bluebird75:
no feedback ... let's merge !

PR #66, opened 2016-07-25
New features to luaunit

ldeniau:
As discussed with Philippe Fremy by email, I do a pull request for features I needed and proved to be extremely useful to test codes, including LuaJIT itself and scientific codes.

Modified:
+ assertTrue/False to be strict (essential), user can still do lazy False (i.e. not x instead of x == false)
+ removed margin in assertAlmostEquals, as it prevents any serious use in numerical codes (its main purpose, right?). Moreover, it is _not_ expected from user point of view to have it's own margin being modified in the back. Without this margin, assertAlmostEquals is much more useful, and can be used  different ways.
Example of use of assertAlmost that don't work with a margin:
assertAlmostEquals( sqrt(v)*sqrt(v) / v - 1, 0, eps )  -- relative error, large scale
assertAlmostEquals( sin(v) - -sin(-v), 0, eps )  -- absolute error
where
eps = 2.2204460492503130e-16  -- double precision epsilon (see C99 standard).

Added:
+ assertIsNaN and variants (true check for numerical NaN)
+ -x, --exclude PATTERN as Lua patterns are not expressive to negate matching. Works as -p.
+ -c, --count NUM to run NUM times tests and ensure JITed code to be tested too (essential with LuaJIT).

Typical use of extensions assuming that Test_XXX are performance tests (longer):
./luajit testsuite.lua -x Test_      # only regression tests
./luajit testsuite.lua -p Test_      # only performance tests
./luajit testsuite.lua -c 4 -x Test_      # test referential transparency of regression tests
./luajit testsuite.lua -c 100 -x Test_      # test JITed version of regression tests

Last use is essential to check consistency between interpreted and compiled code with LuaJIT. I found few bugs and inconsistencies with this feature!

Laurent.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -19,7 +19,7 @@ files = {
         ignore = {"test%w+"}
     },
     ["test/compat_luaunit_v2x.lua"] = {
-        ignore = {"EXPORT_ASSERT_TO_GLOBALS", "[Tt]est[%w_]+", "assert%w+"}
+        ignore = {"EXPORT_ASSERT_TO_GLOBALS", "[Tt]est[%w_]+", "assert[%w_]+"}
     },
     ["test/legacy_example_with_luaunit.lua"] = {
         ignore = {"LuaUnit", "EXPORT_ASSERT_TO_GLOBALS", "[Tt]est[%w_]+",
--- a/README.md
+++ b/README.md
@@ -30,6 +30,10 @@ It is released under the BSD license.
 Documentation is available on
 [read-the-docs](http://luaunit.readthedocs.org/en/latest/)
 
+**Important note when upgrading to version 3.1 and above** : break of backward compatibility, assertions functions are
+no longer exported directly to the global namespace. See [documentation](http://luaunit.readthedocs.io/en/latest/#luaunit-global-asserts) on how to adjust or restore previous behavior.
+
+
 ##Install
 
 The version in development on github is always stable and can be used safely.
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ You can also install it with bower :
 
 **LuaRocks**
 
-A rocks is being worked on.
+LuaUnit v3.2 and above are available on LuaRocks.
 
 ##Contributors
 * [NiteHawk](https://github.com/n1tehawk)
--- a/README.md
+++ b/README.md
@@ -121,6 +121,9 @@ Unofficial fork from version 1.3
 - run() may also be called Run()
 - table deep comparision (also available in 1.4)
 - control verbosity with setVerbosity() SetVerbosity() and set_verbosity()
+- More assertions: 
+    + is<Type>, is_<type>, assert<Type> and assert_<type> (e.g. assert( LuaUnit.isString( getString() ) )
+    + assertNot<Type> and assert_not_<type>
 
 #### Version 1.5 - 8. Nov 2012
 - compatibility with Lua 5.1 and 5.2
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ A rocks is being worked on.
 
 #### Version 3.2 - 12. Jul 2016
 * distinguish between failures (failed assertion) and errors
+* add command-line option to stop on first error or failure
 * support for new versions: Lua 5.3 and LuaJIT (2.0, 2.1 beta)
 * validation of all lua versions on Travis CI and AppVeyor
 * added compatibility layer with forked luaunit v2.x
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -46,6 +46,13 @@ https://github.com/bluebird75/luaunit
 
 It is released under the BSD license.
 
+
+Upgrade note
+================
+
+**Important note when upgrading to version 3.1 and above** : there is a break of backward compatibility in version 3.1, assertions functions are no longer exported directly to the global namespace. See :ref:`luaunit-global-asserts` on how to adjust or restore previous behavior.
+
+
 LuaUnit development
 ===================
 
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -57,6 +57,7 @@ This documentation describes the functionality of LuaUnit v3.2 .
 
 New in version 3.2 - 12. Jul 2016
 ---------------------------------
+* Add command-line option to stop on first error or failure
 * Distinguish between failures (failed assertion) and errors
 * Support for new versions: Lua 5.3 and LuaJIT (2.0, 2.1 beta)
 * Validation of all lua versions on Travis CI and AppVeyor
@@ -225,7 +226,13 @@ You always have:
 * the date at which the test suite was started
 * the group to which the function belongs (usually, the name of the function table, and *<TestFunctions>* for all direct test functions)
 * the name of the function being executed
-* a report at the end, with number of executed test, number of non selected tests, number of failures and duration.
+* a report at the end, with number of executed test, number of non selected tests, number of failures, number of errors (if any) and duration.
+
+The difference between failures and errors are:
+
+* luaunit assertion functions generate failures
+* any unexpected error during execution generates an error
+* failures or errors during setup() or teardown() always generate errors
 
 
 You also want to test that when the function receives negative numbers, it generates an error. Use
@@ -668,6 +675,12 @@ actually match the pattern are selected. When specifying more than one pattern,
 they are tried one by one until the name matches (OR combination).
 
 Make sure you esape magic chars like ``+?-*`` with ``%`` .
+
+**Stopping on first error or failure**
+
+If --failure or -f is passed as an option, LuaUnit will stop on the first failure or error and display the test results.
+
+If --error or -e is passed as an option, LuaUnit will stop on the first error (but continue on failures).
 
 
 **Other Options:**
--- a/luaunit-3.2.1-1.rockspec
+++ b/luaunit-3.2.1-1.rockspec
@@ -1,10 +1,10 @@
 package = "LuaUnit"
-version = "3.2-1"
+version = "3.2.1-1"
 source =
 {
 	url = "git://github.com/bluebird75/luaunit",
 	
-	tag = "LUAUNIT_V3_2"
+	tag = "LUAUNIT_V3_2_1"
 }
 
 description =
@@ -19,9 +19,11 @@ description =
 		platforms (Jenkins, Maven, ...) . The integrated command-line options provide a flexible interface to select tests by name 
 		or patterns, control output format, set verbosity, ...
 
-		LuaUnit works with Lua 5.1, 5.2 and 5.3 . It was tested on Windows XP, Windows Server 2012 R2 (x64) and 
-		Ubuntu 14.04 (see continuous build results on Travis-CI and AppVeyor ) and should work on all platforms 
-		supported by lua. It has no other dependency than lua itself. 
+		LuaUnit works with Lua 5.1, LuaJIT 2.0, LuaJIT 2.1 beta, Lua 5.2 and Lua 5.3 . It is tested on Windows Seven, Windows Server 2012 R2 (x64) and Ubuntu 14.04 (see continuous build results on Travis-CI and AppVeyor) and should work on all platforms supported by Lua.
+		It has no other dependency than Lua itself. 
+
+		**Important note when upgrading to version 3.1 and above** : break of backward compatibility, assertions functions are
+		no longer exported directly to the global namespace. See [documentation](http://luaunit.readthedocs.io/en/latest/#luaunit-global-asserts) on how to adjust or restore previous behavior.
 
 	]],
 	homepage = "http://github.com/bluebird75/luaunit",
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -18,7 +18,7 @@ M.private = {}
 M.VERSION='3.2'
 M._VERSION=M.VERSION -- For LuaUnit v2 compatibility
 
---[[ Some people like assertEquals( actual, expected ) and some people prefer 
+--[[ Some people like assertEquals( actual, expected ) and some people prefer
 assertEquals( expected, actual ).
 ]]--
 M.ORDER_ACTUAL_EXPECTED = true
@@ -53,7 +53,11 @@ Options:
   -o, --output OUTPUT:    Set output type to OUTPUT
                           Possible values: text, tap, junit, nil
   -n, --name NAME:        For junit only, mandatory name of xml file
+  -c, --count NUM:        Execute all tests NUM times, e.g. to trig the JIT
   -p, --pattern PATTERN:  Execute all test names matching the Lua PATTERN
+                          May be repeated to include severals patterns
+                          Make sure you escape magic chars like +? with %
+  -x, --exclude PATTERN:  Exclude all test names matching the Lua PATTERN
                           May be repeated to include severals patterns
                           Make sure you escape magic chars like +? with %
   testname1, testname2, ... : tests to run in the form of testFunction,
@@ -663,13 +667,13 @@ function M.assertError(f, ...)
 end
 
 function M.assertTrue(value)
-    if not value then
+    if value ~= true then
         failure("expected: true, actual: " ..prettystr(value), 2)
     end
 end
 
 function M.assertFalse(value)
-    if value then
+    if value ~= false then
         failure("expected: false, actual: " ..prettystr(value), 2)
     end
 end
@@ -686,6 +690,18 @@ function M.assertNotIsNil(value)
     end
 end
 
+function M.assertIsNaN(value)
+    if type(value) ~= "number" or value == value then
+        failure("expected: nan, actual: " ..prettystr(value), 2)
+    end
+end
+
+function M.assertNotIsNaN(value)
+    if type(value) == "number" and value ~= value then
+        failure("expected non nan value, received nan", 2)
+    end
+end
+
 function M.assertEquals(actual, expected)
     if type(actual) == 'table' and type(expected) == 'table' then
         if not _is_table_equals(actual, expected) then
@@ -698,11 +714,7 @@ function M.assertEquals(actual, expected)
     end
 end
 
--- Help Lua in corner cases like almostEquals(1.1, 1.0, 0.1), which by default
--- may not work. We need to give margin a small boost; EPSILON defines the
--- default value to use for this:
-local EPSILON = 1E-11
-function M.almostEquals( actual, expected, margin, margin_boost )
+function M.almostEquals( actual, expected, margin )
     if type(actual) ~= 'number' or type(expected) ~= 'number' or type(margin) ~= 'number' then
         error_fmt(3, 'almostEquals: must supply only number arguments.\nArguments supplied: %s, %s, %s',
             prettystr(actual), prettystr(expected), prettystr(margin))
@@ -710,8 +722,7 @@ function M.almostEquals( actual, expected, margin, margin_boost )
     if margin < 0 then
         error('almostEquals: margin must not be negative, current value is ' .. margin, 3)
     end
-    local realmargin = margin + (margin_boost or EPSILON)
-    return math.abs(expected - actual) <= realmargin
+    return math.abs(expected - actual) <= margin
 end
 
 function M.assertAlmostEquals( actual, expected, margin )
@@ -986,6 +997,7 @@ local list_of_funcs = {
     { 'assertIsTable'           , 'assert_is_table' },
     { 'assertIsBoolean'         , 'assert_is_boolean' },
     { 'assertIsNil'             , 'assert_is_nil' },
+    { 'assertIsNaN'             , 'assert_is_nan' },
     { 'assertIsFunction'        , 'assert_is_function' },
     { 'assertIsThread'          , 'assert_is_thread' },
     { 'assertIsUserdata'        , 'assert_is_userdata' },
@@ -996,6 +1008,7 @@ local list_of_funcs = {
     { 'assertIsTable'           , 'assertTable' },
     { 'assertIsBoolean'         , 'assertBoolean' },
     { 'assertIsNil'             , 'assertNil' },
+    { 'assertIsNaN'             , 'assertNaN' },
     { 'assertIsFunction'        , 'assertFunction' },
     { 'assertIsThread'          , 'assertThread' },
     { 'assertIsUserdata'        , 'assertUserdata' },
@@ -1006,6 +1019,7 @@ local list_of_funcs = {
     { 'assertIsTable'           , 'assert_table' },
     { 'assertIsBoolean'         , 'assert_boolean' },
     { 'assertIsNil'             , 'assert_nil' },
+    { 'assertIsNaN'             , 'assert_nan' },
     { 'assertIsFunction'        , 'assert_function' },
     { 'assertIsThread'          , 'assert_thread' },
     { 'assertIsUserdata'        , 'assert_userdata' },
@@ -1016,6 +1030,7 @@ local list_of_funcs = {
     { 'assertNotIsTable'        , 'assert_not_is_table' },
     { 'assertNotIsBoolean'      , 'assert_not_is_boolean' },
     { 'assertNotIsNil'          , 'assert_not_is_nil' },
+    { 'assertNotIsNaN'          , 'assert_not_is_nan' },
     { 'assertNotIsFunction'     , 'assert_not_is_function' },
     { 'assertNotIsThread'       , 'assert_not_is_thread' },
     { 'assertNotIsUserdata'     , 'assert_not_is_userdata' },
@@ -1026,6 +1041,7 @@ local list_of_funcs = {
     { 'assertNotIsTable'        , 'assertNotTable' },
     { 'assertNotIsBoolean'      , 'assertNotBoolean' },
     { 'assertNotIsNil'          , 'assertNotNil' },
+    { 'assertNotIsNaN'          , 'assertNotNaN' },
     { 'assertNotIsFunction'     , 'assertNotFunction' },
     { 'assertNotIsThread'       , 'assertNotThread' },
     { 'assertNotIsUserdata'     , 'assertNotUserdata' },
@@ -1036,6 +1052,7 @@ local list_of_funcs = {
     { 'assertNotIsTable'        , 'assert_not_table' },
     { 'assertNotIsBoolean'      , 'assert_not_boolean' },
     { 'assertNotIsNil'          , 'assert_not_nil' },
+    { 'assertNotIsNaN'          , 'assert_not_nan' },
     { 'assertNotIsFunction'     , 'assert_not_function' },
     { 'assertNotIsThread'       , 'assert_not_thread' },
     { 'assertNotIsUserdata'     , 'assert_not_userdata' },
@@ -1267,7 +1284,7 @@ then OK or FAILED (failures=1, error=1)
 Started
  .
  Finished in 0.002695 seconds.
- 
+
  1 tests, 2 assertions, 0 failures, 0 errors
 
 -- Ruby:
@@ -1276,13 +1293,13 @@ Loaded suite tc_simple_number2
 Started
 F..
 Finished in 0.038617 seconds.
- 
+
   1) Failure:
 test_failure(TestSimpleNumber) [tc_simple_number2.rb:16]:
 Adding doesn't work.
 <3> expected but was
 <4>.
- 
+
 3 tests, 4 assertions, 1 failures, 0 errors
 
 -- Java Junit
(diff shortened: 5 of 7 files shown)


ldeniau:
'margin' in my comment should be understood as 'realmargin' in the code where
realmargin = margin + (margin_boost or EPSILON)
so the message is "do not modify user's margin"

The pattern I used for performance tests was Test_ because it was easier to exclude them before the --exclude option was added, e.g. -p Test[^_] to avoid them. But with --exclude one can use proper naming like TestPerf prefix, and -x TestPerf and -p TestPerf to exclude or include the performance tests respectively.

Laurent.

n1tehawk:
Personally, I am biased **against** the proposed change in `assertTrue`/`assertFalse`. Why?

Currently, these two functions closely reflect an aspect of Lua's language philosophy: boolean expressions and operators will consider anything as `true` that is not explicitly `false` or `nil`. If you wish to do a "strict" test for logical constants, this is readily available with `assertEquals`:
```lua
local lu = require("luaunit")

local t, f = "yes", nil

lu.assertTrue(t)
--lu.assertEquals(t, true) -- FAILS
lu.assertFalse(f)
--lu.assertEquals(f, false) -- FAILS
```
My impression is we would sacrifice the "proper" (= Lua's own) evaluation of the test expression, and gain no real advantage over using `assertEquals`. To me that means a loss of flexibility.

n1tehawk:
Your criticism of `assert[Not]AlmostEquals` is certainly valid. LuaUnit uses a default EPSILON that's rather large, which easily "covers up" small numerical errors.

Actually this was precisely the reason why I introduced the optional `margin_boost` parameter in commit f9569dcf. The `or` operator in [luaunit.lua#L713](https://github.com/bluebird75/luaunit/blob/3836cd1ba6502c9364d6ea485e71033c23faa183/luaunit.lua#L713) means that this parameter - if specified - would "win" over the default epsilon, including the case where the user would request a zero `margin_boost`. Unfortunately, this was implemented only half-heartedly, with the current versions of `assertAlmostEquals` and `assertNotAlmostEquals` not being able to use/pass a user-provided margin boost.

We need a certain tradeoff (= default epsilon) between numerical correctness and Lua-internal rounding errors that might prevent these functions from delivering the expected results. Our 'classic' test case is `assertEquals(1.1, 1, 0.1)` which actually _fails_ with no additional "boost".

I haven't re-thought this in depth yet, but my impression holds that a viable strategy is to have a way (e.g. an optional parameter) of "overriding" or cancelling the default boost value.

bluebird75:
First, thanks Laurent to go through the effort of the Pull Request and community review. It is never a quiet journey. And even more thanks to sharing with us your experience of scientific computing.

I have several comments : 

1. assertIsNan / assertIsNotNan : no problem to include this assertion. Can you elaborate which lua package gives you these types ? I would like to document them properly. There is also some tests needed to validate them but I can take care of it myself. Is there any other similar non standard types which are worth adding ?

2. assertTrue / assertIsTrue : I believe there is no definite answer on what is the correct behavior. People with different background will have different expecation between _should use lua way of computing true/false_ or _should assert strictly to true/false values_ . I like @n1tehawk  suggestion of assertIsTrue which fits more in the spirit on how luaunit is organising its asserts. The key is anyway to provide good documentation on this debate and add notes to avoid people falling into the pitfall.

3. exclude patterns : I like the idea  .  I will play with it and add more tests to see if we can combine include/exclude more intelligently.

4. Repeating tests : I like the idea a lot as well. This would be first step in customizing test execution. I actually have more ideas on this topic.

5. assertAlmostEquals : clearly, luaunit's current approach is weak and unfit to scientific computing. Imrovements are needed. What do you think about the margin_boost parameter ?

Also, I noticed that in mad, you define : https://github.com/MethodicalAcceleratorDesign/MAD/blob/master/tests/utests/luacore.lua#L37 eps and tiny and you favour eps for your margins rather than tiny. Any reason ?

tiny is the smallest 64 bits double. Can you elaborate on the value eof eps ? It would be interesting for us to have test cases where linux/windows portability for example forces a different margin on a given test...

n1tehawk:
>tiny is the smallest 64 bits double. Can you elaborate on the value eof eps ? It would be interesting for us to have test cases where linux/windows portability for example forces a different margin on a given test...

`eps` is equal to 2^(-52) in that case, which represents the [machine epsilon](https://en.wikipedia.org/wiki/Machine_epsilon) for "double precision" numbers.

ldeniau:
> On 25 Jul 2016, at 20:50, NiteHawk <notifications@github.com> wrote:
> 
> Personally, I am biased against the proposed change in assertTrue/assertFalse. Why?
> 
> Currently, these two functions closely reflect an aspect of Lua's language philosophy: boolean expressions and operators will consider anything as true that is not explicitly false or nil.
> 
You are right for boolean expression, this is arguable for assertion since both are possible. The goal of unit tests is to be as strict as possible.
> If you wish to do a "strict" test for logical constants, this is readily available with assertEquals:
> 
> local lu = require("luaunit")
> 
> local t, f = "yes", nil
> 
> lu.assertTrue(t)
> --lu.assertEquals(t, true) -- FAILS
> lu.assertFalse(f)
> --lu.assertEquals(f, false) -- FAILS
> My impression is we would sacrifice the "proper" (= Lua's own) evaluation of the test expression, and gain no real advantage over using assertEquals. To me that means a loss of flexibility.
> 
In fact both can emulate the other, wether you prefer
assertEquals(f, false) is equivalent to strict assertFalse(f) (i.e. proposed)
and
assertTrue(not f) is equivalent to non-strict assertFalse(f) (i.e. actual)

My proposal is not against lua philosophy, it just explicits more clearly the user’s intent. For the same reason, I like assertNil(x) vs assertEquals(x, nil) because the intend is clearer and strict.

assertTrue(not f) clearly states that we accept false and nil, while assertTrue(t) clearly states that we expect true (and nothing else).
while with non-strict assertTrue(t) and assertFalse(f), the user can be confused by the misleading names, i.e. one would expect t==true and f==false respectively.

My suggestion was just to avoid a different internal logic from the naming, as well as providing unit tests as strict as possible. It is very easy to miss a bug with lazy tests:

t = { false, false, nil, false, false }
for i=1,#t do
	assertFalse(t[i])  — actual non-strict false.
end
— test passed, #t == 5, KO consider invalid sequence as valid sequence.

t = { false, false, false, nil, false }
for i=1,#t do
	assertFalse(t[i])  — actual non-strict false.
end
— test passed, #t == 3, OK sequence truncated (implementation defined)

This unit test is in fact unpredictable unless you know how the implementation check for #t…. With strict assertFalse, first test will failed and detect invalid Lua sequence.

So as a general statement, it is always better to be as strict as possible when simple alternatives can be used to explicitly relax the constraints: assertTrue(not x) is clear to me.

I hope that I have clarified why I proposed to be as strict as possible as far as it does not reduce the flexibility. Even then, lazy behaviour should be considered with care, especially in the critical field of unit testing!  It is already difficult enough to validate code in dynamically typed languages, let’s not add more lazy behaviour in assertions...

Best,
Laurent.

PS. I have be confused my by both, extra margin and non-strict assertFalse, and I lost one day (in total) to investigate why my tests were not _failing_ as expected. The first guess being usually that you do something wrong and check your code base _before_ looking the luaunit implementation.

bluebird75:
I read more stuff on Nan. So no package needed, it is builtin if you have IEEE compliant float handling, which is often the case, but not always for example with gcc specific options.

The stackoverflow answer is interesting :
http://stackoverflow.com/questions/12102222/how-to-test-for-1-ind-indeterminate-in-lua

I learnt more about float computation today :-)

ldeniau:
> On 25 Jul 2016, at 22:25, NiteHawk <notifications@github.com> wrote:
> 
> Your criticism of assert[Not]AlmostEquals is certainly valid. LuaUnit uses a default EPSILON that's rather large, which easily "covers up" small numerical errors.
> 
> Actually this was precisely the reason why I introduced the optional margin_boost parameter in commit f9569dc <https://github.com/bluebird75/luaunit/commit/f9569dcf6f49584d9e8e51aecc9dd26eaee1164e>. The or operator in luaunit.lua#L713 <https://github.com/bluebird75/luaunit/blob/3836cd1ba6502c9364d6ea485e71033c23faa183/luaunit.lua#L713> means that this parameter - if specified - would "win" over the default epsilon, including the case where the user would request a zero margin_boost. Unfortunately, this was implemented only half-heartedly, with the current versions of assertAlmostEquals and assertNotAlmostEquals not being able to use/pass a user-provided margin boost.
> 
> We need a certain tradeoff (= default epsilon) between numerical correctness and Lua-internal rounding errors that might prevent these functions from delivering the expected results. Our 'classic' test case is assertEquals(1.1, 1, 0.1) which actually fails with no additional "boost".
> 
> I haven't re-thought this in depth yet, but my impression holds that a viable strategy is to have a way (e.g. an optional parameter) of "overriding" or cancelling the default boost value.
> 
I agree that without any user provided epsilon, something like 10^-12 could be a good alternative (must be clearly documented), but user provided margins should be absolutely used untouched. Your case above is not equal because 0.1 is a rational number in base 2, so you will have an error typically of one or two 2^-52 (eps). I have other example where you can have an avalanche effect between two rational numbers (speaking in binary) that creates 40 or 80 eps errors. I am not sure that it is possible to have such effect in the range of 10^-12. For example, math.deg(pi/3)-60 created an error of 32*eps, same for 1.6*0.1 (two rational numbers).

Best,
Laurent.

ldeniau:
Hi,

NiteHawk already answered to some points.

> On 25 Jul 2016, at 23:40, Bluebird75 <notifications@github.com> wrote:
> 
> First, thanks Laurent to go through the effort of the Pull Request and community review. It is never a quiet journey. And even more thanks to sharing with us your experience of scientific computing.
> 
> I have several comments :
> 
> assertIsNan / assertIsNotNan : no problem to include this assertion.
> 
Few implementations are possible. The one I proposed is the fastest, and I don’t know any architecture where it would fails (even non IEEE 754 compliant should fail on x==x when x is a NaN)
> Can you elaborate which lua package gives you these types ?
> 
print(0/0)
> I would like to document them properly. There is also some tests needed to validate them but I can take care of it myself. Is there any other similar non standard types which are worth adding ?
> 
the potentially other useful one for numeric would be infinity

assertIsInf(x)   -> if abs(x) ~= 0/1 then “fails” end

but it’s also easy to do it with assertEquals (like all others ;-) 
> assertTrue / assertIsTrue : I believe there is no definite answer on what is the correct behavior. People with different background will have different expecation between should use lua way of computing true/false or should assert strictly to true/false values . I like @n1tehawk <https://github.com/n1tehawk> suggestion of assertIsTrue which fits more in the spirit on how luaunit is organising its asserts. The key is anyway to provide good documentation on this debate and add notes to avoid people falling into the pitfall.
> 
Right. But I would go for what people would guess in the absence of documentation and looking at the API of LuaUnit. Personally, if the strict version is not retained, I will probably never use assertTrue or assertFalse, and I will favour assertEquals systematically. I understand the philosophy of Lua, but unit testing is a serious field ;-) In the same vein, some unit tests found a bug in 

For example, this initialisation (meta)method returns either a or __init(a), where __init is a user’s defined function that can finalise the initialisation process of an object (like __new but for post processing of built object, not preprocessing of constructor arguments):

local function init (a)
  local m = rawget(getmetatable(a), '__init')
  return m and m(a)
end

but as being part of object construction, it must never return nil, always a or a modified version of a. Do you see the bug?
> exclude patterns : I like the idea . I will play with it and add more tests to see if we can combine include/exclude more intelligently.
> 
In fact, they are already combined the right way I think, excluded pattern filter-out already selected tests, either all or by --pattern. The opposite order of combining filters does not make sense to me.
> Repeating tests : I like the idea a lot as well. This would be first step in customizing test execution. I actually have more ideas on this topic.
> 
In fact I started to pass the iteration number by argument to the tests, in case the user need specific behaviour based on iteration number. It was a mistake: referential transparency is a key property of unit tests, this is why everything should go through setUp and tearDown themselves called inside the loop to guarantee to preserve this property. Parametric tests is an error to my humble opinion and were never introduced in TDD for good reason I think. It is too tempting to play with parameters and break this essential property by accident.
> assertAlmostEquals : clearly, luaunit's current approach is weak and unfit to scientific computing. Imrovements are needed. What do you think about the margin_boost parameter ?
> 
I think that something like

M.EPSILON=10^-12
function M.almostEquals( actual, expected, margin )
    margin = margin or M.EPSILON
    if type(actual) ~= 'number' or type(expected) ~= 'number' or type(margin) ~= 'number' then
        error_fmt(3, 'almostEquals: must supply only number arguments.\nArguments supplied: %s, %s, %s',
            prettystr(actual), prettystr(expected), prettystr(margin))
    end
    if margin < 0 then
        error('almostEquals: margin must not be negative, current value is ' .. margin, 3)
    end
    return math.abs(expected - actual) <= margin
end

should be perfect for all cases, including user defined EPSILON, tiny error test, relative error tests, etc... The tests will automatically ensure proper margin or proper M.EPSILON (in case user would change it).
> Also, I noticed that in mad, you define : https://github.com/MethodicalAcceleratorDesign/MAD/blob/master/tests/utests/luacore.lua#L37 <https://github.com/MethodicalAcceleratorDesign/MAD/blob/master/tests/utests/luacore.lua#L37> eps and tiny and you favour eps for your margins rather than tiny. Any reason ?
> 
tiny is the smallest representable value for double precision normalised floating point numbers (i.e. with denormalised number deactivated and automatically rounded to zero, recommended for numerical stability and performance). huge is the biggest representable value. eps is the smallest *relative* error between two double precision floating point numbers (the minimum to distinguish them), it is 2^-52 (number of bits in mantissa, excluding the abstract leading 1). You can also use it as the smallest absolute error between two double precision floating point numbers around 1 (where absolute error ~ relative error)...

see DBL_MIN, DBL_MAX and DBL_EPSILON in C standard from float.h
http://en.cppreference.com/w/c/types/limits
> tiny is the smallest 64 bits double. Can you elaborate on the value eof eps ? It would be interesting for us to have test cases where linux/windows portability for example forces a different margin on a given test...
> 
It’s not windows or linux that matters but the compiler math options (setup when program or library or vm are loaded). Some language/libraries are known to modify these FPU word control when you load them (e.g. Java VM). So one has to ensure proper FPU setUP. LuaJIT rely on SSE2 FPU, not x87 FPU on Intel CPU for good reasons! 

See the following link for a fun bug in octave:
http://octave.1599824.n4.nabble.com/yet-another-Java-oddity-td4649593.html


Best,
Laurent.

ldeniau:
> On 25 Jul 2016, at 23:58, Bluebird75 <notifications@github.com> wrote:
> 
> I read more stuff on Nan. So no package needed, it is builtin if you have IEEE compliant float handling, which is often the case, but not always for example with gcc specific options.
> 
> The stackoverflow answer is interesting :
> http://stackoverflow.com/questions/12102222/how-to-test-for-1-ind-indeterminate-in-lua <http://stackoverflow.com/questions/12102222/how-to-test-for-1-ind-indeterminate-in-lua>
> I learnt more about float computation today :-)
> 
Information in this thread is incomplete. for example 1/0 is not a NaN but +inf…
Beware that LuaJIT use “nan” as a NaN but it is a string, not a number, which is not “standard”. That is why my proposal is checking for a number first. Another implementations as portable as my proposal, but probably slower:
local s = tostring(x)
if type(x) == “number” and (s == “nan” or s == “NaN”) then "it’s a nan" end

see
http://lua-users.org/wiki/InfAndNanComparisons
except that huge (C DBL_MAX) is not Inf on IEEE 754 compliant architecture (unless setup for that if possible). 

Best,
Laurent.

n1tehawk:
Hi Laurent!

>I think that something like
>
>M.EPSILON=10^-12
>function M.almostEquals( actual, expected, margin )
>    margin = margin or M.EPSILON
>    --[[ ... ]]
>    return math.abs(expected - actual) <= margin
>end
>
>should be perfect for all cases, including user defined EPSILON, tiny error test, relative error tests, etc... The tests will automatically ensure proper margin or proper M.EPSILON (in case user would change it).

We need to stay compatible with the existing `assert[Not]Equals` logic that already requires three distinct parameters with the meaning of "`a` differs from `b` by no more than `margin`". To compensate for inaccuracies, IMHO we need to add a fourth parameter on top of that (either implicitly or explicitly), which would reflect the epsilon (or "boost").

Given that we would expose the `margin_boost` to the user-side functions (and with the existing implementation of `almostEquals`), one could do:
```lua
local lu = require("luaunit")

local eps = 2.220446049250313e-16
lu.assertAlmostEquals(1.1, 1, 0.1) -- PASSES (default "boost")
lu.assertAlmostEquals(1.1, 1, 0.1, eps) -- PASSES (user-defined epsilon)
lu.assertAlmostEquals(1.1, 1, 0.1, 0) -- zero "boost", FAILS due to rounding errors
```

ldeniau:
Hi,

After looking at my late reply yesterday night, I think that definitely assertTrue is not strict enough for unit testing, and really misleading the user. It will succeed for anything else than nil and false, a rather dangerous behaviour!

If the intend is to have assertion behaving as in an “if" test, I would name them assertIs(x) and assertIsNot(t) which reflect more what the test is really doing.

I did these modifications because current assertFalse and assertTrue were _dangerous_ for my unit tests (and I guess for other users too). Even with good documentation, users will be tempted to use assertTrue without care, trusting it for true test. And assertTrue(x == true) or assertEquals(x, true) does not seems to be in the philosophy of unit testing, otherwise assertEquals would be enough for everything (except NaN and error). And IMHO unit test philosophy prevails over Lua philosophy in the context of unit testing.

Best,
Laurent.

ldeniau:
Hi,

> On 26 Jul 2016, at 09:35, NiteHawk <notifications@github.com> wrote:
> 
> Hi Laurent!
> 
> I think that something like
> 
> M.EPSILON=10^-12
> function M.almostEquals( actual, expected, margin )
> margin = margin or M.EPSILON
> --[[ ... ]]
> return math.abs(expected - actual) <= margin
> end
> 
> should be perfect for all cases, including user defined EPSILON, tiny error test, relative error tests, etc... The tests will automatically ensure proper margin or proper M.EPSILON (in case user would change it).
> 
> We need to stay compatible with the existing assert[Not]Equals logic that already requires three distinct parameters with the meaning of "a differs from b by no more than margin". To compensate for inaccuracies, IMHO we need to add a fourth parameter on top of that (either implicitly or explicitly), which would reflect the epsilon (or "boost").
> 
> Given that we would expose the margin_boost to the user-side functions (and with the existing implementation of almostEquals), one could do:
> 
> local lu = require("luaunit")
> 
> local eps = 2.220446049250313e-16
> lu.assertAlmostEquals(1.1, 1, 0.1) -- PASSES (default "boost")
> lu.assertAlmostEquals(1.1, 1, 0.1, eps) -- PASSES (user-defined epsilon)
> lu.assertAlmostEquals(1.1, 1, 0.1, 0) -- zero "boost", FAILS due to rounding errors
This approach is not suitable at all for numerical codes, you cannot check relative errors (more than half of my tests)… As I said, assertAlmostEquals should either provide a default epsilon for common non numerical use, and should not modify the users margin when provided. Otherwise, believe me, it will become useless for serious use. See in
https://github.com/MethodicalAcceleratorDesign/MAD/blob/master/tests/utests/gmath.mad
how I am using it (207 times), at least in four different ways that are needed depending on the scale of the numerics and the range of the errors: absolute error vs expected, absolute error minus expected versus zero, relative error vs expected, relative error minus expected versus one. All these usages are important, only one is possible as soon as you modify the user margin…

You seems to be more attached to Lua philosophy than to unit testing philosophy. When users spend more than half of their time writing tests, they prefer a framework that is as strict as possible, rock solid, and cristal clear about the semantic. Convention, modification in the back or lazy semantic should be absolutely prohibited. If right now you need to break some backward compatibility, do it. In the context of unit testing, users will thank you for strictness, rigour, robustness an simplicity of your framework and does not care about smart default arguments. On the other hand, if the framework is ambiguous, or not rock solid, it will not use it for testing his code. Breaking backward compatibility is not an issue if you fulfilled the points above, you already did it two weeks ago for table comparison. I had tests that was passing, that suddenly failed with the new version, with a net result that I found a bug in my object model! I will always thank you to help me detect bugs with LuaUnit, but I will stop use it (or keep my modifications) if it doesn’t detect bugs...

Best,
Laurent.

n1tehawk:
>If the intend is to have assertion behaving as in an “if" test, I would name them assertIs(x) and assertIsNot(t) which reflect more what the test is really doing.

That should probably read "assertI**f**" and "assertNotI**f**"? (the latter is more in line with LuaUnit's naming conventions than "assertI**f**Not")

BTW: This pull request is a bit ill-prepared and should be rebased / refactored. You have some unrelated commit and merge history that really doesn't belong here; the only commit introducing any actual changes is https://github.com/bluebird75/luaunit/pull/66/commits/6e5f7b59e60f34bf0d2086f20e6e7849dc784c5a. It would also be nice if that commit could be split up into smaller parts, following the practice of "one commit per logical change".

Regards, NiteHawk

ldeniau:
Hi,

> On 26 Jul 2016, at 10:03, NiteHawk <notifications@github.com> wrote:
> 
> If the intend is to have assertion behaving as in an “if" test, I would name them assertIs(x) and assertIsNot(t) which reflect more what the test is really doing.
> 
> That should probably read "assertIf" and "assertNotIf"? (the latter is more in line with LuaUnit's naming conventions than "assertIfNot")
> 
Fine! I don’t have preferences with naming, and following your convention is of course the best.
> BTW: This pull request is a bit ill-prepared and should be rebased / refactored. You have some unrelated commit and merge history that really doesn't belong here; the only commit introducing any actual changes is 6e5f7b5 <https://github.com/bluebird75/luaunit/pull/66/commits/6e5f7b59e60f34bf0d2086f20e6e7849dc784c5a>. It would also be nice if that commit could be split up into smaller parts, following the practice of "one commit per logical change".
> 
Well, this is my first pull request and I am rather new to “advanced" stuff with git (I usually do patches, I am a bit old-style), so I certainly mixed up few unrelated things…

How can I improve the situation? I guess that I should rebase my branch and add/commit step by step the changes. I will give another try. What is the luaunit commit where I should add my changes, because I suspect that I merged (by accident) commit that are ahead of the master, right?

Cheers,
Laurent.

n1tehawk:
>How can I improve the situation? I guess that I should rebase my branch and add/commit step by step the changes. I will give another try. What is the luaunit commit where I should add my changes, because I suspect that I merged (by accident) commit that are ahead of the master, right?

The problem seems to be that the _master_ branch in [your fork](https://github.com/MethodicalAcceleratorDesign/luaunit) has deviated from "upstream" several commits before the change that you wanted to end up in the pull request, with the common "ancestor" being 3836cd1. I'm not quite sure what happened after that, but it seems that you (first?) pulled in your local changes (leading to a "split" in commit history) and then tried to catch up with _upstream/master_ afterwards. But as a PR will always contain everything back to the common ancestry, you ended up with a lot of "unrelated" (and redundant) commits.

To remedy the situation I think you should first create a "backup" branch, preserving the current commit history. This will also allow you to cherry-pick the relevant commit 87da910 whenever it is needed again. After that you can try to fix your current _master_ to reflect the upstream status more closely again (this probably involves `git reset` to an earlier commit and then merging / fast-forwarding from _upstream_ again). Then you may re-add (= cherry-pick) 87da910, or an equivalent set of patches that break down the changes into smaller units. Finish by force-pushing the new _master_, which in turn should update this PR accordingly. If you need help, let me know and I'll see what I can do to assist you.

(Alternatively, you could of course also abandon this PR and start a fresh one.)

Regards, NiteHawk

ldeniau:
> On 26 Jul 2016, at 11:37, NiteHawk <notifications@github.com> wrote:
> 
> How can I improve the situation? I guess that I should rebase my branch and add/commit step by step the changes. I will give another try. What is the luaunit commit where I should add my changes, because I suspect that I merged (by accident) commit that are ahead of the master, right?
> 
> The problem seems to be that the master branch in your fork <https://github.com/MethodicalAcceleratorDesign/luaunit> has deviated from "upstream" several commits before the change that you wanted to end up in the pull request, with the common "ancestor" being 3836cd1 <https://github.com/bluebird75/luaunit/commit/3836cd1ba6502c9364d6ea485e71033c23faa183>. I'm not quite sure what happened after that, but it seems that you (first?) pulled in your local changes (leading to a "split" in commit history) and then tried to catch up with upstream/master afterwards. But as a PR will always contain everything back to the common ancestry, you ended up with a lot of "unrelated" (and redundant) commits.
> 
> To remedy the situation I think you should first create a "backup" branch, preserving the current commit history. This will also allow you to cherry-pick the relevant commit 87da910 <https://github.com/bluebird75/luaunit/commit/87da910d069a5b69bf5e07d32e3b22d9d7dfa627> whenever it is needed again. After that you can try to fix your current master to reflect the upstream status more closely again (this probably involves git reset to an earlier commit and then merging / fast-forwarding from upstream again). Then you may re-add (= cherry-pick) 87da910 <https://github.com/bluebird75/luaunit/commit/87da910d069a5b69bf5e07d32e3b22d9d7dfa627>, or an equivalent set of patches that break down the changes into smaller units. Finish by force-pushing the new master, which in turn should update this PR accordingly. If you need help, let me know and I'll see what I can do to assist you.
> 
> (Alternatively, you could of course also abandon this PR and start a fresh one.)
> 
> 

This is what I did (just finished), because it was better to split the changes in dedicated commits.

So I created a new branch and added step by step my proposed changes with better commit comments so you can trace each steps. If added assertIf and assertNotIf as you recommended (see the message log below)

$ git branch
* ld_feature_newtry
  master


$ git status
On branch ld_feature_newtry
nothing to commit, working directory clean

$ git log
commit b728380bf7368a4f053e71a8948ccd5046215f2f
Author: Laurent Deniau <laurent.deniau@cern.ch>
Date:   Tue Jul 26 11:44:52 2016 +0200

    assertIsInf and assertIsNotInf and variants added

commit 248da00fc0ba9dccf3ef82c65ed2f3a2cc8dce79
Author: Laurent Deniau <laurent.deniau@cern.ch>
Date:   Tue Jul 26 11:37:16 2016 +0200

    Remove 'local' to EPSILON definition (typo)

commit eaf335a1d8826a0692d3537dad31d75f13ae5650
Author: Laurent Deniau <laurent.deniau@cern.ch>
Date:   Tue Jul 26 11:35:31 2016 +0200

    Modify (simplify) almostEquals to make it more flexible and usable for testing numeric codes, still keep EPSILON as default margin to help common cases, leave user margin untouched if provided (main concern!)

commit 9c867a90107d75b68e5313c9f3fe1297a9daaa1c
Author: Laurent Deniau <laurent.deniau@cern.ch>
Date:   Tue Jul 26 11:29:54 2016 +0200

    assertTrue and assertFalse made strict, non-strict version renamed assertIf and assertNotIf respectively (and variant added)

commit 6b00239da9689dba615324503322f67cc7183dcc
Author: Laurent Deniau <laurent.deniau@cern.ch>
Date:   Tue Jul 26 11:20:25 2016 +0200

    assertIsNaN and assertIsNotNaN and variants added

commit f8efb3ebf6a9a6c6289f6e6d48ce40ff8e183c06
Author: Laurent Deniau <laurent.deniau@cern.ch>
Date:   Tue Jul 26 11:13:45 2016 +0200

    -c, --count num added to run num times each test and allow to test its JITed version (e.g. num = 100). Also useful (with lower num, e.g. num = 4) to check reference transparency of tests, a key property of unit tests.

commit f4f11e831af40ff0e2b3aa4d902116800998713e
Author: Laurent Deniau <laurent.deniau@cern.ch>
Date:   Tue Jul 26 11:06:16 2016 +0200

    -x, --exclude pattern added to allow easier pattern negation for tests selection, not so easy with Lua patterns

commit 9c6ae90a0f8016c5d49735a0633a9c3b51b2f1f3
Author: Philippe F <phil@freehackers.org>
Date:   Wed Jul 20 08:24:16 2016 +0200

    Make compatibility break more explicit


So the next step I guess is to push the commit on my fork, but it is probably where I selected wrong branch (something ahead of upstream/master, merging other development  branches to mine during the push).

What are the next safe steps? Thanks for your help.

Best,
Laurent.

ldeniau:
What about

git push --set-upstream origin ld_feature_newtry

and then pull a request?

n1tehawk:
Hi Laurent!

As I understand it, `ld_feature` and `ld_feature_newtry` are your _local_ branches that you did not push to github (your fork only has a _master_ branch)?

One possibility would be to publish (= push) that new branch, and create a new PR from it - as you have suggested right now. :smile:

ldeniau:
Here my local situation (GitUp graph):

ldeniau:
> On 26 Jul 2016, at 12:12, NiteHawk <notifications@github.com> wrote:
> 
> Hi Laurent!
> 
> As I understand it, ld_feature and ld_feature_newtry are your local branches that you did not push to github (your fork only has a master branch)?
> 
> One possibility would be to publish (= push) that new branch, and create a new PR from it - as you have suggested right now. 😄
> 
the problem is that push does not work and suggests the command I just sent:

git push
fatal: The current branch ld_feature_newtry has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin ld_feature_newtry


And considering the graph of GitUp, I don’t want to screw-up my branch or other branches, so which one should be taken? I would go for upstream/master, but this also what I did this night, I guess...

Cheers,
Laurent.

n1tehawk:
`git push --set-upstream origin ld_feature_newtry` should be fine, and won't affect this pull request as long as you don't modify your existing _master_. Once you have published your branch, it should be easier to check on it and decide on the next steps.

Issue #67, opened 2016-07-25
Run luaunit on lua compiled with 32bits numbers


bluebird75:
It would be interesting to see if we have some 64bits number assumptions in luaunit. For what I know, compiling lua with 32bits floats is common in the embedded world. Worth checking !

n1tehawk:
I've started some experiments with this in my [20160726_float32 branch](https://github.com/n1tehawk/luaunit/commits/20160726_float32).

The Lua source distribution contains a dedicated header file `luaconf.h` to allow for some customization. One of the settings that can be tweaked is the C type used for `LUA_NUMBER`. So it's actually pretty easy to set this to `float` (32-bit) instead of the default `double`.

On the other hand, Lua**JIT** explicitly warns against changing these defaults (see [luaconf.h#L117-L119](https://github.com/LuaJIT/LuaJIT/blob/db1b399af134e715b8ba9a0e4ad9cbafbe06ecb1/src/luaconf.h#L117-L119)) - as it is rather sensitive to having a `lua_Number` type of sufficiently large size. When you insist on changing it to `float`, LuaJIT will crash miserably - e.g. segfaults on Travis CI, see https://travis-ci.org/n1tehawk/luaunit/builds/147455077.

We can safely extend the build matrix and the Lua build steps to use single-precision numbers. Without further adjustments, this leads to a number of formatting errors however: https://travis-ci.org/n1tehawk/luaunit/builds/147461067. The problem is that the original `LUA_NUMBER_FMT` output format is no longer appropriate and needs to be adjusted.

After doing so, there's only one test that remains failing for all of the Lua 5.x builds: https://travis-ci.org/n1tehawk/luaunit/builds/147463747. Unsurprisingly, that's `TestLuaUnitAssertions.test_assertAlmostEquals` - caused by the fact that the rounding errors (due to loss of precision) now exceed our EPSILON adjustment for _margin_.

n1tehawk:
I've just notices that Lua 5.2 seems to have problems with this approach. I'll have to investigate that more closely with a local copy.

bluebird75:
I think testing in regular lua should be sufficient. The idea is to catch the limitaions of 32bits float arithmetic.

We should probably add a bit more or arithmetic tests verify typical usage of assertAlmostEquals.

n1tehawk:
I'll clean up the branch and convert it into a pull request, so we can start testing with 32-bit floats then.

bluebird75:
Done !

PR #69, opened 2016-07-27
Extend Travis CI to test "single" precision Lua configurations

n1tehawk:
This pull request follows up on the suggestion (in #67) to test Lua configurations with 32-bit floats. It adds three new entries to the build matrix that will use customized versions of Lua (5.1, 5.2 and 5.3) where the `LUA_NUMBER` type gets modified accordingly.


bluebird75:
Nice one again !

PR #71, opened 2016-08-02
Suggest alternative approach for dealing with numerical assertions

n1tehawk:
This pull request is related to ideas introduced in #66 and #68. @ldeniau pointed out that our existing `almostEquals()` function has some shortcomings when it comes to 'scientific' testing of numerical errors / precision.

To address these concerns, some changes to `almostEquals()` have been suggested - most notably the ability to prevent any modification of the user-provided error limit ("margin"). I've been rethinking this over the weekend and would like to take it another step further: Introduce a new set of functions that deal with "numerical error assertions".

- `assertAbsErrorWithin()` and `assertNotAbsErrorWithin()` test for _absolute_ error/margin
- `assertRelErrorWithin()` and `assertNotRelErrorWithin()` test for _relative_ error/margin

I'd like your opinion on this.

Regards, NiteHawk

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -24,6 +24,7 @@ assertEquals( expected, actual ).
 M.ORDER_ACTUAL_EXPECTED = true
 M.PRINT_TABLE_REF_IN_ERROR_MSG = false
 M.TABLE_EQUALS_KEYBYCONTENT = true
+M.ALMOST_EQUALS_USES_EPSILON = true
 M.LINE_LENGTH=80
 
 --[[ M.EPSILON is meant to help with Lua's floating point math in simple corner
@@ -721,7 +722,12 @@ function M.almostEquals( actual, expected, margin, margin_boost )
     if margin < 0 then
         error('almostEquals: margin must not be negative, current value is ' .. margin, 3)
     end
-    local realmargin = margin + (margin_boost or M.EPSILON)
+    local realmargin
+    if M.ALMOST_EQUALS_USES_EPSILON then
+        realmargin = margin + (margin_boost or M.EPSILON)
+    else
+        realmargin = margin + (margin_boost or 0)
+    end
     return math.abs(expected - actual) <= realmargin
 end
 
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -26,6 +26,22 @@ M.PRINT_TABLE_REF_IN_ERROR_MSG = false
 M.TABLE_EQUALS_KEYBYCONTENT = true
 M.LINE_LENGTH=80
 
+--[[ M.EPSILON is meant to help with Lua's floating point math in simple corner
+cases like almostEquals(1.1-0.1, 1), which may not work as-is (e.g. on numbers
+with rational binary representation) if the user doesn't provide some explicit
+error margin.
+The default margin used by almostEquals() in such cases is M.EPSILON; and since
+Lua may be compiled with different numeric precisions (single vs. double), we
+try to select a useful default for it dynamically. Note: If the initial value
+is not acceptable, it can be changed by the user to better suit specific needs.
+See also: https://en.wikipedia.org/wiki/Machine_epsilon
+]]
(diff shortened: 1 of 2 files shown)


ldeniau:
Hi,

> On 02 Aug 2016, at 10:41, NiteHawk <notifications@github.com> wrote:
> 
> This pull request is related to ideas introduced in #66 <https://github.com/bluebird75/luaunit/pull/66> and #68 <https://github.com/bluebird75/luaunit/pull/68>. @ldeniau <https://github.com/ldeniau> pointed out that our existing almostEquals() function has some shortcomings when it comes to 'scientific' testing of numerical errors / precision.
> 
> To address these concerns, some changes to almostEquals() have been suggested - most notably the ability to prevent any modification of the user-provided error limit ("margin"). I've been rethinking this over the weekend and would like to take it another step further: Introduce a new set of functions that deal with "numerical error assertions".
> 
> assertAbsErrorWithin() and assertNotAbsErrorWithin() test for absolute error/margin
> assertRelErrorWithin() and assertNotRelErrorWithin() test for relative error/margin

As already discussed, there is a dozen of useful relative errors and the choice depends on the range of the numbers and the precision you want on the error check. In this field, the user must be educated whatever you propose, and therefore assertAlmostEqual as proposed should be enough.

Example if possible relative errors to be injected in absolute error check:

|(a-b)|			vs n*eps*|a|
|(a-b)|			vs n*eps*|b|
|(a-b) / a|		vs n*eps 
|(a-b) / b|		vs n*eps
 |a-b| / min(|a|,|b|)	vs n*eps
 |a-b| / max(|a|,|b|)	vs n*eps
 |a/b|			vs n*eps+1
 |b/a|			vs n*eps+1
 |a/b - 1|		vs n*eps
 |b/a - 1|		vs n*eps

Then you have all the asymmetric signed variants where two bounds are needed, like
eps_low <= (a-b) / min(|a|,|b|) <= eps_high

So to keep it simple, the only one you should provide is the one we discussed and proposed (absolute error with user margin _or_ default small margin). If you want absolutely to propose another one, provide something like

assertAlmostEqualsWithin(expected, actual, low_margin, high_margin)
  ensure low_margin <= high_margin and low_margin <= expected-actual <= high_margin

The more advanced you could provide (not recommended!) is something that adjust the margin to the number of significant digits, something like margin = eps * 10^(-max(#a,#b)), where eps is the user margin on accepted deviation of significant digits and #a and #b is the number of significant digits in the number (need to detect leading/trailing 0). This problem is non-trivial (see https://cds.cern.ch/record/1574594/files/CERN-ACC-NOTE-2013-0005.pdf).

But keep assertions simple, user writing unit tests must understand what their tests are checking. As for non-strict assertions, over-designed assertions are _dangerous_. Instead of providing a bunch of over-designed assertions, keep assertAlmostEquals as proposed and provide some examples of use in the documentation. This will be much easier to maintain and to use. The root of most references for floating point arithmetic is the paper from Golberg, "What Every Computer Scientist Should Know About Floating-Point Arithmetic” available everywhere on the web, could be also mentioned in the documentation.
> I'd like your opinion on this.
> 
> Regards, NiteHawk
> 

Best,
Laurent.

n1tehawk:
This is obviously more complicated than I had anticipated. Thanks for setting things straight!

Regards, NiteHawk

ldeniau:
You’re welcome. What about?

assertAlmostEqualsWithin(expected, actual, low_margin, high_margin)
  ensure low_margin <= high_margin and low_margin <= expected-actual <= high_margin

This is simple, and just as useful as assertAlmostEquals. Maybe rename it to something shorter, e.g.
assertEqualsWithin(expected, actual, low_margin, high_margin)
could be enough (if not already used). Then you could even think of implementing 
assertAlmostEquals as a special case of assertEqualsWithin
where -low_margin == high_margin == margin.
I can update my pull request for that.

Best,
Laurent.

PR #72, opened 2016-08-02
stricter assertTrue/False & assertEvalToTrue/False

bluebird75:
Here is my proposal (along with unit-tests) for stricter assertTrue/False.

I propose also the name assertEvalToTrue/False which tells precisely what it does.

@ldeniau, @n1tehawk, what do you think ?


n1tehawk:
Looks good to me.

doc/index.rst
```diff
@@ -771,6 +771,31 @@ not influence the test itself.
 
 Value assertions
 ----------------------
+
+LuaUnit contains several flavours of true/false assertions, to be used in different contexts.
+Usually, when asserting for true or false, you want strict assertions (nil should not 
+assert to false); assertTrue and assertFalse are the functions for this purpose. In some cases though,
+you want Lua coerction rules to apply (value 1, string "hello" yields true) and the right functions to use
```
n1tehawk: There's a typo here: coerction &rarr; coercion

PR #73, opened 2016-08-05
Randomize test order

bluebird75:
For review.

I was thinking about randomising only the test table content, to avoid mixing methods of different table but full random looks simpler and nicer.

And I was happy to see no side effects on luaunit testing...

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -176,18 +176,16 @@ local function sortedPairs(tbl)
 end
 M.private.sortedPairs = sortedPairs
 
-math.randomseed(os.clock()*100000000000)
+-- seed the random with a strongly varying seed
+math.randomseed(os.clock()*1E11)
 
 local function randomizeTable( t )
-    -- Shuffle the table t
-    local rand = math.random 
-    assert( t, "shuffleTable() expected a table, got nil" )
-    local iterations = #t
-    local j
-    
-    for i = iterations, 2, -1 do
-        j = rand(i)
-        t[i], t[j] = t[j], t[i]
+    -- randomize the item orders of the table t
+    for i = #t, 2, -1 do
+        local j = math.random(i)
+        if i ~= j then
+            t[i], t[j] = t[j], t[i]
+        end
     end
 end
 M.private.randomizeTable = randomizeTable
@@ -2096,7 +2094,7 @@ end
 
         local expandedList = self.expandClasses( listOfNameAndInst )
         if self.randomize then
-            M.private.randomizeTable( expandedList )
+            randomizeTable( expandedList )
         end
         local filteredList, filteredOutList
             = self.applyPatternFilter( self.patternFilter, expandedList )
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -50,6 +50,7 @@ Options:
   -q, --quiet:            Set verbosity to minimum
   -e, --error:            Stop on first error
   -f, --failure:          Stop on first failure or error
+  -r, --random            Run tests in random order
   -o, --output OUTPUT:    Set output type to OUTPUT
                           Possible values: text, tap, junit, nil
   -n, --name NAME:        For junit only, mandatory name of xml file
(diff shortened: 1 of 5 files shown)


luaunit.lua
```diff
@@ -175,6 +175,22 @@ local function sortedPairs(tbl)
 end
 M.private.sortedPairs = sortedPairs
 
+math.randomseed(os.clock()*100000000000)
```
n1tehawk: I find this hard to read - what about `math.randomseed(os.clock() * 1E11)`?

luaunit.lua
```diff
@@ -175,6 +175,22 @@ local function sortedPairs(tbl)
...
+
+local function randomizeTable( t )
+    -- Shuffle the table t
+    local rand = math.random 
+    assert( t, "shuffleTable() expected a table, got nil" )
```
n1tehawk: 1. I guess that you have probably changed the function name - but it now mismatches the assertion, which says "shuffleTable()".
2. If you want a table type, then test for it - `assert(t)` may be insufficient:

```lua
    if type(t) ~= "table" then
        error("randomizeTable() expected a table, got " .. type(t))
    end
```

luaunit.lua
```diff
@@ -175,6 +175,22 @@ local function sortedPairs(tbl)
...
+    local iterations = #t
+    local j
+    
+    for i = iterations, 2, -1 do
+        j = rand(i)
+        t[i], t[j] = t[j], t[i]
+    end
```
n1tehawk: We could cut down on `local`s here. I don't think that "caching" `math.random` makes much of a difference, and `iterations` is only used to initialize the loop variable, so I would do away with the `local rand` and suggest

```lua
    for i = #t, 2, -1 do
        local j = math.random(i)
        t[i], t[j] = t[j], t[i]
    end
```

luaunit.lua
```diff
@@ -2074,6 +2094,9 @@ end
         ]]
 
         local expandedList = self.expandClasses( listOfNameAndInst )
+        if self.randomize then
+            M.private.randomizeTable( expandedList )
```
n1tehawk: The function is in the local scope, so there's no need to take the "private" route here - simply call `randomizeTable(expandedList)`.

test/test_luaunit.lua
```diff
@@ -138,6 +138,23 @@ TestLuaUnitUtilities = { __class__ = 'TestLuaUnitUtilities' }
...
     end
 
+    function TestLuaUnitUtilities:test_randomizeTable()
+        local t, tref, n, i
+        t, tref, n, i = {}, {}, 20, 0
+        while i < n do
+            table.insert(t, i)
+            table.insert(tref, i)
+            i = i + 1
+        end
```
n1tehawk:
```lua
        local t, tref, n = {}, {}, 20
        for i = 0, n - 1 do
            table.insert(t, i)
            table.insert(tref, i)
        end
```

luaunit.lua
n1tehawk: Optional: `math.random(i)` may return `i`, so maybe it would be useful to avoid the swapping if `j == i`?
```lua
        if j ~= i then
            t[i], t[j] = t[j], t[i]
        end
```

test/test_luaunit.lua
n1tehawk: We could even avoid the `table.insert` easily by using [1, n] instead of [0, n-1]:
```lua
        local t, tref, n = {}, {}, 20
        for i = 1, n do
            t[i], tref[i] = i, i
        end
```

bluebird75:
Thanks for the insightful comments. I'll resubmit...

n1tehawk:
I like the new version better now, thanks!

You have dropped the (table) type check for `t` in `randomizeTable()`, was that on purpose? It shouldn't be harmful as that is a 'private' function (internal to LuaUnit), and we expect only invocations that are controlled by us.

You have kept a coding style on `test_randomizeTable()` that's slightly more verbose than what I would use. But that's a matter of personal taste, and nothing to be concerned about.

I have no objections against merging this PR now.

Regards, NiteHawk

PR #76, opened 2016-10-09
Mathematical tests for "not a number" (NaN) and "infinite", as proposed by ldeniau

n1tehawk:
(Plus unit tests added by me,)

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -789,6 +789,18 @@ function M.assertNotIsNaN(value)
     end
 end
 
+function M.assertIsInf(value)
+    if type(value) ~= "number" or math.abs(value) ~= 1/0 then
+        failure("expected: inf, actual: " ..prettystr(value), 2)
+    end
+end
+
+function M.assertNotIsInf(value)
+    if type(value) == "number" and math.abs(value) == 1/0 then
+        failure("expected non inf value, received ±inf", 2)
+    end
+end
+
 function M.assertEquals(actual, expected)
     if type(actual) == 'table' and type(expected) == 'table' then
         if not _is_table_equals(actual, expected) then
@@ -1097,6 +1109,7 @@ local list_of_funcs = {
     { 'assertIsTrue'            , 'assert_is_true' },
     { 'assertIsFalse'           , 'assert_is_false' },
     { 'assertIsNaN'             , 'assert_is_nan' },
+    { 'assertIsInf'             , 'assert_is_inf' },
     { 'assertIsFunction'        , 'assert_is_function' },
     { 'assertIsThread'          , 'assert_is_thread' },
     { 'assertIsUserdata'        , 'assert_is_userdata' },
@@ -1110,6 +1123,7 @@ local list_of_funcs = {
     { 'assertIsTrue'            , 'assertTrue' },
     { 'assertIsFalse'           , 'assertFalse' },
     { 'assertIsNaN'             , 'assertNaN' },
+    { 'assertIsInf'             , 'assertInf' },
     { 'assertIsFunction'        , 'assertFunction' },
     { 'assertIsThread'          , 'assertThread' },
     { 'assertIsUserdata'        , 'assertUserdata' },
@@ -1123,6 +1137,7 @@ local list_of_funcs = {
     { 'assertIsTrue'            , 'assert_true' },
     { 'assertIsFalse'           , 'assert_false' },
     { 'assertIsNaN'             , 'assert_nan' },
+    { 'assertIsInf'             , 'assert_inf' },
     { 'assertIsFunction'        , 'assert_function' },
     { 'assertIsThread'          , 'assert_thread' },
     { 'assertIsUserdata'        , 'assert_userdata' },
@@ -1136,6 +1151,7 @@ local list_of_funcs = {
     { 'assertNotIsTrue'         , 'assert_not_is_true' },
     { 'assertNotIsFalse'        , 'assert_not_is_false' },
     { 'assertNotIsNaN'          , 'assert_not_is_nan' },
+    { 'assertNotIsInf'          , 'assert_not_is_inf' },
     { 'assertNotIsFunction'     , 'assert_not_is_function' },
     { 'assertNotIsThread'       , 'assert_not_is_thread' },
     { 'assertNotIsUserdata'     , 'assert_not_is_userdata' },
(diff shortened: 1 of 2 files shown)


n1tehawk:
Philippe, I am assuming that you plan to integrate the assertTrue/False changes (#72) eventually?
Regards, NiteHawk

test/test_luaunit.lua
```diff
@@ -1035,6 +1035,33 @@ TestLuaUnitAssertions = { __class__ = 'TestLuaUnitAssertions' }
...
 
+    function TestLuaUnitAssertions:test_assertIsNaN()
+        assertFailure(lu.assertIsNaN, "hi there!")
+        assertFailure(lu.assertIsNaN, nil)
+        assertFailure(lu.assertIsNaN, {})
+        assertFailure(lu.assertIsNaN, {1,2,3})
+        assertFailure(lu.assertIsNaN, {1})
+        assertFailure(lu.assertIsNaN, coroutine.create( function(v) local y=v+1 end ) )
+        lu.assertIsNaN(0 / 0)
+        assertFailure(lu.assertIsNaN, 0 / 1) -- 0.0
```
bluebird75: Let's validate nan with a few more use cases. You can pick from : http://stackoverflow.com/questions/25506281/what-are-all-the-possible-calculations-that-could-cause-a-nan-in-python

test/test_luaunit.lua
```diff
@@ -1137,6 +1164,33 @@ TestLuaUnitAssertions = { __class__ = 'TestLuaUnitAssertions' }
...
 
+    function TestLuaUnitAssertions:test_assertNotIsNaN()
+        lu.assertNotIsNaN( "hi there!" )
+        lu.assertNotIsNaN( nil )
+        lu.assertNotIsNaN( {} )
+        lu.assertNotIsNaN( {1,2,3} )
+        lu.assertNotIsNaN( {1} )
+        lu.assertNotIsNaN( coroutine.create( function(v) local y=v+1 end ) )
+        assertFailure(lu.assertNotIsNaN, 0 / 0)
+        lu.assertNotIsNaN( 0 / 1 ) -- 0.0
```
bluebird75: Same remark as above, let's validate with a few more test cases, just for the sake of completeness.

luaunit.lua
```diff
@@ -753,6 +753,30 @@ function M.assertNotIsNil(value)
...
+
+function M.assertNotIsNaN(value)
+    if type(value) == "number" and value ~= value then
+        failure("expected non nan value, received nan", 2)
+    end
+end
+
+function M.assertIsInf(value)
+    if type(value) ~= "number" or math.abs(value) ~= 1/0 then
+        failure("expected: inf, actual: " ..prettystr(value), 2)
```
bluebird75: Don't you think that it would be better to use a constant for comparing with inf. I am uneasy with performing the calculation each time...

ldeniau:
Hi,

What do you mean by a separate test?

isInf is supposed to identify if a number is an infinity or not. Getting the sign is like for any number, i.e. <0 or >0. For signed infinity, assertEquals does the job.

I often use

assertEquals ( 1/f(), -inf ) to test if f() returns -0 or +0, so in this case I need to capture the sign. assertIsInf is when you don’t care about the sign.

Best,
Laurent.

luaunit.lua
n1tehawk: It should be possible to use `math.huge` for that - see e.g. http://lua-users.org/wiki/InfAndNanComparisons. The unit tests I provided did check that `math.huge` is "infinite".

bluebird75:
Laurent, do you see any need for a function such as assertPlusInf() and assertMinusInf() ?

I like also your +0 / -0 test, is it something which would make sense for you in an assertion library with good support for mathematics computing ... like luaunit could become ? :-)

n1tehawk:
I have added two more commits to address some remarks from above. Travis CI test passed (https://travis-ci.org/n1tehawk/luaunit/builds/166683021)

ldeniau:
Hi,

> On 11 Oct 2016, at 11:04, Bluebird75 <notifications@github.com> wrote:
> 
> Laurent, do you see any need for a function such as assertPlusInf() and assertMinusInf() ?
> 
No, as I said, Equals does the job.
> I like also your +0 / -0 test, is it something which would make sense for you in an assertion library with good support for mathematics computing ... like luaunit could become ? :-)
> 
Could be useful for someone that don’t know how to do.

assertPositiveZero
assertNegativeZero

but if you want to automatise such test, then it must check for zero first to avoid NaN...

assertPositiveZero:
assert( x == 0 and 1/x ==  infinity)

assertNegativeZero:
assert( x == 0 and 1/x == -infinity)

Best,
Laurent.

ldeniau:
> On 11 Oct 2016, at 11:10, Laurent Deniau <laurent.deniau@gmail.com> wrote:
> 
> Hi,
> 
>> On 11 Oct 2016, at 11:04, Bluebird75 <notifications@github.com <mailto:notifications@github.com>> wrote:
>> 
>> Laurent, do you see any need for a function such as assertPlusInf() and assertMinusInf() ?
>> 
> No, as I said, Equals does the job.
>> I like also your +0 / -0 test, is it something which would make sense for you in an assertion library with good support for mathematics computing ... like luaunit could become ? :-)
>> 
> Could be useful for someone that don’t know how to do.
> 
> assertPositiveZero
> assertNegativeZero
> 
> but if you want to automatise such test, then it must check for zero first to avoid NaN...
> 
> assertPositiveZero:
> assert( x == 0 and 1/x ==  infinity)
> 
> assertNegativeZero:
> assert( x == 0 and 1/x == -infinity)
> 

Sorry, better to test it this way

assertPositiveZero:
assert(type(x) == “number” and x == 0 and 1/x == infinity)

assertNegativeZero:
assert(type(x) == “number” and x == 0 and 1/x == -infinity)

because there is no problem with NaN here. The test for == 0 is needed because you might end up with an infinity with very small values close to zero.

That is why in my application, I introduced tiny and huge (~= infinity) values as the smallest/largest finite representable numbers to avoid all the confusion encountered in C/C++ and unfortunately Lua by extension. Generating tiny and huge is difficult (better to get them as constants somewhere), while infinity and NaN are trivial...

Best,
Laurent.

bluebird75:
@n1tehawk : yes, please rebase.

n1tehawk:
Rebased to b7a00c1.

PR #77, opened 2016-10-21
Some minor changes and fixups

n1tehawk:
- be clearer on M.EPSILON
- whitespace and code formatting
- remove some unnecessary code from _is_table_equals()
- update AppVeyor to use Lua 5.3.3


n1tehawk:
Somehow github's CI integration hooks didn't pick up on this PR. I'll try to force-push my branch to get it to 'catch up' again...

bluebird75:
The commit is fine, I will accept it.

However, after re-reading the discussion with Laurent, I think he makes a good point that luaunit should not introduce a bias in the calculation. marginBoost should go away...

I'll discuss it further in #68

PR #78, opened 2016-11-08
Better list diffing

bluebird75:
Improved difference analysis between lists of more than 10 elements.


bluebird75:
I'll resubmitwith a clearer separation between regulqr fixups and new code

bluebird75:
Cleaner version available here: #79

PR #79, opened 2016-11-10
Better list diffing2

bluebird75:
Cleaned up version of the list diff analysis


run_functional_tests.lua
```diff
@@ -146,11 +148,11 @@ local function adjustFile( fileOut, fileIn, pattern, mayBeAbsent, verbose )
             if verbose then
                 print('Modifying line: '..line )
             end
-            line = string_sub(line, dest, source)
+            line = string_gsub(line, dest, source)
             -- line = line:sub(1,idxStart-1)..source..line:sub(idxEnd+1)
             -- string.gsub( line, dest, source )
             if verbose then
-                print('Result: '..line )
+                print('Result        : '..line )
```
n1tehawk: That whitespace looks weird, is it really supposed to be there?

n1tehawk: Ah, I got it. It brings the line's prefix up to the same length as `'Modifying line: '` further up, for aligment / easier comparison.

bluebird75:
When I use a local, it does not work. I get the following error message : 

    d:\work\luaunit\luaunit-git\luaunit2>lua run_unit_tests.lua
    is_equalsfunction: 0083B6F8`
    .....................................................................................is_equalsnil
    F....................................is_equalsnil
    E......................is_equalsnil
    E...
    Failed tests:

    1) TestLuaUnitErrorMsg.test_printTableWithRef

    .\test\test_luaunit.lua:1937: Error message does not contain: "lists <table: "

    Error message received: ".\luaunit.lua:661: attempt to call global 'is_equals' (a nil value)"


    stack traceback:
        .\test\test_luaunit.lua:1937: in function 'TestLuaUnitErrorMsg.test_printTableWithRef'


    2) TestLuaUnitUtilities.test_diffAnalysisThreshold

    .\luaunit.lua:661: attempt to call global 'is_equals' (a nil value)

    stack traceback:
        .\test\test_luaunit.lua:253: in function 'TestLuaUnitUtilities.test_diffAnalysisThreshold'


    3) TestLuaUnitUtilities.test_suitableForMismatchFormatting

    .\luaunit.lua:661: attempt to call global 'is_equals' (a nil value)

    stack traceback:
        .\test\test_luaunit.lua:246: in function 'TestLuaUnitUtilities.test_suitableForMismatchFormatting'


    Ran 149 tests in 0.024 seconds, 146 successes, 1 failure, 2 errors

The spurious prints qre because I added the extra lines to debug the situation :

lines 658-661 :

    i = 1
    print( 'is_equals'..tostring(is_equals ) )
    while i <= longest do
         if not is_equals(ta[i], tb[i]) then

This is the is_equals which triggers the error.

lines 1044-1047 :

    M.private._is_table_equals = _is_table_equals
    local is_equals = _is_table_equals
    print( 'is_equals'..tostring(is_equals ) )

bluebird75:
The build fails because lua does not have the same representation of numbers on windows and linux...

Annoying for test based on textual representations :-)

n1tehawk:
Okay, I now see where the problem lies: Since your new routines are further up in the source code, this is a case of "use before declare" - leading to those errors.

There are two possible solutions: Either a) move the entire `_is_table_equals()` block/logic further up (or your list formatting further down); or b) declare something like
```lua
local is_equals -- alias for _is_table_equals() function, gets assigned below
```
near the top of the file (e.g. under "general utility functions"), and then do a simple `is_equals = _is_table_equals` after the function has been declared.

Regards, NiteHawk

n1tehawk:
>The build fails because lua does not have the same representation of numbers on windows and linux...

We have encountered and worked around number formatting issues before, e.g. [test/test_luaunit.lua#L1671-L1675](
https://github.com/bluebird75/luaunit/blob/9c43106535a3bfcdf2e14dcbbbed21a8dbbc3efc/test/test_luaunit.lua#L1671-L1675).

Maybe this would be a good opportunity to tackle the problem at the `prettystr()` level, eliminating the differences between Lua versions?

n1tehawk:
https://github.com/n1tehawk/luaunit/commit/e0d1d188452c387aa4a06c4cde42191fb998d789 illustrates what I have in mind...

bluebird75:
That's actually a good idea ! We can do it after merging in this branch.

You want to take care of it ?

Apart from that, any further comments on this branch ?

n1tehawk:
>That's actually a good idea ! We can do it after merging in this branch.
>You want to take care of it ?

Sure, I can do that.

>Apart from that, any further comments on this branch ?

Not really. I've noted a few smaller things here and there, but it's a bit hard to keep track of this PR in its entirety (as it needs some rebasing onto the current master by now). I think you should merge it first, if specific issues come up after that, we can still improve on / fix them later.

If a more general _luacheck_ is giving you troubles (makes you struggle with Lua syntax), I'll gladly help with that too.

Regards, NiteHawk

PR #80, opened 2016-11-18
Add a new outputter class that supports ANSI-colored text

n1tehawk:
This is eye-candy based on the existing "text" class, augmented by ANSI control sequences for different colors.

To use it, request "color" output, e.g.:
`./run_functional_tests.lua -v --output color`


n1tehawk:
Will rebase onto current master...

bluebird75:
Initially, I was thinking about a command line option --color to deal with color. However, I see that busted also used an outputter approach to deal with the color thing, so why not...

However, the text output itself should be identical with and without colors. Internally, the best way to go is to make this an option of the general text outputting class.

I did not check your branch under windows yet, but as far as I know, unix color codes don't work in the default windows console.

[Googling a bit on the topic]

So, on windows 10, ansi color codes work out of the box. On previous versions of windows, using program such as ansicon can make them work but I find the approach tedious.

So, as first step, let's just enable colors on unix. We will improve later on Windows support.

n1tehawk:
I selected the "outputter" class as a natural approach at first, but I do agree that we're duplicating code from the text output. And I assume that we don't expect to require any other output that would be 'colored'?

Re-uniting the code should be straightforward enough. We could start with a COLOR table that initially holds empty strings, and only set those (to the needed control sequences) when `--color` is encountered. If older Windows (Win32 console) needs special treatment, we could introduce a different option (`--wincolor`?) to adjust the control strings. The text outputter would just always concatenate the COLOR strings to its output; in the case of "plain" text those would simply be empty.

n1tehawk:
Fixed typo in commit message, rebased to current `master`

PR #81, opened 2016-11-18
Some fixes, uniform prettystr() output for numbers

n1tehawk:
(see https://github.com/bluebird75/luaunit/pull/79#issuecomment-260209580)


bluebird75:
good, merging in.

PR #82, opened 2016-11-27
Combine include and exclude

bluebird75:
Combine together include and exclude pattern to allow for complex test list selections

### DIFF
--- a/test/test_luaunit.lua
+++ b/test/test_luaunit.lua
@@ -566,9 +566,10 @@ TestLuaUnitUtilities = { __class__ = 'TestLuaUnitUtilities' }
         lu.assertEquals( lu.LuaUnit.parseCmdLine( { '-p', 'toto' } ), { pattern={'toto'} } )
         lu.assertEquals( lu.LuaUnit.parseCmdLine( { '-p', 'titi', '-p', 'toto' } ), { pattern={'titi', 'toto'} } )
         lu.assertErrorMsgContains( 'Missing argument after -p', lu.LuaUnit.parseCmdLine, { '-p', } )
-        lu.assertEquals( lu.LuaUnit.parseCmdLine( { '--exclude', 'toto' } ), { exclude={'toto'} } )
-        lu.assertEquals( lu.LuaUnit.parseCmdLine( { '-x', 'toto' } ), { exclude={'toto'} } )
-        lu.assertEquals( lu.LuaUnit.parseCmdLine( { '-x', 'titi', '-x', 'toto' } ), { exclude={'titi', 'toto'} } )
+        lu.assertEquals( lu.LuaUnit.parseCmdLine( { '--exclude', 'toto' } ), { pattern={'!toto'} } )
+        lu.assertEquals( lu.LuaUnit.parseCmdLine( { '-x', 'toto' } ), { pattern={'!toto'} } )
+        lu.assertEquals( lu.LuaUnit.parseCmdLine( { '-x', 'titi', '-x', 'toto' } ), { pattern={'!titi', '!toto'} } )
+        lu.assertEquals( lu.LuaUnit.parseCmdLine( { '-x', 'titi', '-p', 'foo', '-x', 'toto' } ), { pattern={'!titi', 'foo', '!toto'} } )
         lu.assertErrorMsgContains( 'Missing argument after -x', lu.LuaUnit.parseCmdLine, { '-x', } )
 
         -- count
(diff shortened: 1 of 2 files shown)


n1tehawk:
I like the idea in general. But my impression is that this particular implementation of `patternFilter()` is causing a bit of trouble, and some of the new test cases don't seem very intuitive to me.

The third parameter (`nil_result`) should be removed from the function, as _luaunit.lua_ itself will now never pass anything else but `true` there.

My main concern is that the result of `patternFilter()` now **depends on the order of patterns** passed to the function. In turn that may lead to inconsistent results of `applyPatternFilter()`. To demonstrate this, add these lines at the end of `TestLuaUnitUtilities:test_applyPatternFilter()`:
```lua
-- test for consistent result on parameter reordering
testset = {
    { 'foo.one', dummy}, { 'foo.two', dummy},
    { 'bar.one', dummy}, { 'bar.two', dummy},
    { 'baz.one', dummy}, { 'baz.two', dummy},
}
included, excluded = lu.LuaUnit.applyPatternFilter( {'ba.', '!one'}, testset )
lu.assertEquals( included, {testset[4], testset[6]} )
lu.assertEquals( #excluded, 4 )
included, excluded = lu.LuaUnit.applyPatternFilter( {'ba.', '!two'}, testset )
lu.assertEquals( included, {testset[3], testset[5]} )
lu.assertEquals( #excluded, 4 )
included, excluded = lu.LuaUnit.applyPatternFilter( {'!one', 'ba.'}, testset )
lu.assertEquals( included, {testset[4], testset[6]} )
lu.assertEquals( #excluded, 4 )
included, excluded = lu.LuaUnit.applyPatternFilter( {'!two', 'ba.'}, testset )
lu.assertEquals( included, {testset[3], testset[5]} )
lu.assertEquals( #excluded, 4 )
included, excluded = lu.LuaUnit.applyPatternFilter( {'two', '!foo', 'one'}, testset )
lu.assertEquals( included, {testset[3], testset[4], testset[5], testset[6]} )
lu.assertEquals( #excluded, 2 )
included, excluded = lu.LuaUnit.applyPatternFilter( {'!foo', 'one', 'two'}, testset )
lu.assertEquals( included, {testset[3], testset[4], testset[5], testset[6]} )
lu.assertEquals( #excluded, 2 )
included, excluded = lu.LuaUnit.applyPatternFilter( {'o..', '..o', '!foo'}, testset )
lu.assertEquals( included, {testset[3], testset[4], testset[5], testset[6]} )
lu.assertEquals( #excluded, 2 )
```

n1tehawk:
To fix this, I would suggest a set of rules that allow `patternFilter()` to act on patterns regardless of their ordering:
* The function should be able to determine the default return value (in case no pattern is matched) on its own. With no patterns at all, or only `--exclude` patterns, the default should be `true`. But users (rightfully) expect to be able to restrict the set of tests by specifying a `--pattern`; so as soon as one or more 'inclusive' patterns are given, we should change to a `false` default return value.
* _Any_ match of an `--exclude` pattern will always mean a negative result (= return `false` immediately) - i.e. exclusions will always "win" over other (inclusive) patterns. The idea is that something like `--pattern f.. --exclude foo` will allows to select a selection of tests, and then exclude specific ones. Given "fun", "foo" and "bar", these options should always select (only) "fun" - independent of the order of options.

With that, I arrive at a pattern filter function that looks like this:
```lua
local function patternFilter(patterns, expr, obsolete)
    assert(obsolete == nil)
    -- Run `expr` through the inclusion and exclusion rules defined in patterns
    -- and return true if expr shall be included, false for excluded.
    -- Inclusion pattern are defined as normal patterns, exclusions 
    -- patterns start with `!`. If patterns is `nil` (= unset) or empty,
    -- or none of the patterns matches expr, return a suitable default.

    -- result: nil = UNKNOWN (not matched yet), true = ACCEPT, false = REJECT
    -- default: true if no explicit "include" is found, set to false otherwise
    local default, result = true, nil

    if patterns ~= nil then
        for _, pattern in ipairs(patterns) do
            local exclude = pattern:sub(1,1) == '!'
            if exclude then
                pattern = pattern:sub(2)
            else
                -- at least one include pattern specified, a match is required
                default = false
            end
            -- print('pattern: ',pattern)
            -- print('exclude: ',exclude)
            -- print('default: ',default)

            if string.find(expr, pattern) then
                if exclude then
                    return false -- reject (an explicit exclude always "wins")
                end
                -- getting here means we've got an "include" match
                result = true -- accept
            end
        end
    end

    -- after processing all patterns, return either result or the default value
    return result or default
end
```
When done with testing, the `obsolete` assertion should be removed for the final version, of course.

test/test_luaunit.lua
```diff
@@ -605,6 +605,27 @@ TestLuaUnitUtilities = { __class__ = 'TestLuaUnitUtilities' }
...
+
+        lu.assertEquals( lu.private.patternFilter( { '!foo' }, 'foo'), false )
+        lu.assertEquals( lu.private.patternFilter( { '!foo', 'foo' }, 'foo'), true )
+        lu.assertEquals( lu.private.patternFilter( { '!foo', 'foo', '!foo' }, 'foo'), false )
+        lu.assertEquals( lu.private.patternFilter( { '!foo', 'foo', '!foo', 'foo' }, 'foo'), true )
+
+        lu.assertEquals( lu.private.patternFilter( { 'f..', '!foo', '__foo__' }, 'toto'), false )
+        lu.assertEquals( lu.private.patternFilter( { 'f..', '!foo', '__foo__' }, 'fii'), true )
+        lu.assertEquals( lu.private.patternFilter( { 'f..', '!foo', '__foo__' }, 'foo'), false )
+        lu.assertEquals( lu.private.patternFilter( { 'f..', '!foo', '__foo__' }, '__foo__'), true )
```
n1tehawk: This corresponds to a case where options were given that say "select `__foo__` and anything matching `f..`, but I don't want stuff that matches `foo`". I'd find it more natural to return `false` here.

It's debatable if a more specific `__foo__` should override the exclusion - but that would come with the big drawback of becoming dependent on the specific order of patterns...

bluebird75:
Making the selection depending on pattern order is precisely one of the goal of this patch. This allows to build complex selection rules.
 
My idea was that include rules are equivalent to `or match(pattern)` and exclusion rules are equivalent to `and not match(pattern)`.

Another way to formulate this is that include rules should add new tests to the selection from the global corpus and exclude rules should remove tests from the existing selection. If the first encountered rules is an include rulem the selection starts empty. If the first encountered rule is an exclusion, the selection set starts with the full corpus.

It seems that this is not a complete success as you pointed out in the following line : 

    lu.assertEquals( lu.private.patternFilter( { '!f..', 'foo', '!__foo__' }, 'toto'), true )

This indeed should return `false` .

I don't get however why you find counter-intuitive that include and exclude rules must be order independant. This allows for less powerful selection rules (although we are certainly talking about a super mintority of use case).

n1tehawk:
Okay, you have a point there. I was probably thinking too much along the lines of "preserve existing options behavior"...

I think you could keep your `result` logic and just add the default (return value) selection that I suggested on top - that should also solve the test case I pointed out.

n1tehawk:
I still think the inner loop in `patternFilter()` can be simplified a lot.
```lua
-- result: nil = UNKNOWN (not matched yet), true = ACCEPT, false = REJECT
-- default: true if no explicit "include" is found, set to false otherwise
local default, result = true, nil

if patterns ~= nil then
    for _, pattern in ipairs(patterns) do
        local exclude = pattern:sub(1,1) == '!'
        if exclude then
            pattern = pattern:sub(2)
        else
            -- at least one include pattern specified, a match is required
            default = false
        end
        -- print('pattern: ',pattern)
        -- print('exclude: ',exclude)
        -- print('default: ',default)

        if string.find(expr, pattern) then
            -- set result to false when excluding, true otherwise
            result = not exclude
        end
    end
end

if result ~= nil then
    return result
end
return default
```

bluebird75:
I'll merge it to have the general feature available, and feel free to improve it.

Issue #85, opened 2016-12-23
add assertions if table contains values


ChristianKuehnel:
Proposal for extension:
As a developer I want to check in my test case if a table does (not) contain a certain value.

It could look like this in the code:
```
table = some_function_to_be_tested(...)
luaunit.assertContains(table, "some value")
luaunit.assertNotContains(table, "other value")
```
This is my current workaround:
```
tablex = require('pl.tablex')
function contains(table, value)
  return tablex.find(table, value) ~= nil
end

luaunit.assertTrue(contains(table,"some value"))
luaunit.assertFalse(contains(table,"other value"))
```

n1tehawk:
Hi Christian!

This implies that you do not know the corresponding table key, correct?
Otherwise you could do `luaunit.assertNotIsNil(t[key])`.

Regards, NiteHawk

ChristianKuehnel:
Hi NiteHawk,

correct. I want to use this mainly for testing arrays and there I do not care about the keys or the order of the values in the array.

Regards, 
Christian

bluebird75:
The request is sound, we'll try to work on something to address this need.

bluebird75:
closed by #87

PR #87, opened 2017-01-24
Implement assertContains() and assertNotContains()

n1tehawk:
as requested in #85

### DIFF
--- a/test/test_luaunit.lua
+++ b/test/test_luaunit.lua
@@ -1233,6 +1233,44 @@ TestLuaUnitAssertions = { __class__ = 'TestLuaUnitAssertions' }
         assertFailure(lu.assertItemsEquals, {one=1,two=2,three=3}, {two=2,one=1})
     end
 
+    function TestLuaUnitAssertions:test_assertTableContains()
+        local t = {3, 'some value', 1, 2}
+
+        assertFailure(lu.assertTableContains, t, 0)
+        lu.assertTableContains(t, 1)
+        lu.assertTableContains(t, 2)
+        lu.assertTableContains(t, 3)
+        assertFailure(lu.assertTableContains, t, 4)
+        lu.assertTableContains(t, 'some value')
+        assertFailure(lu.assertTableContains, t, 'other value')
+
+        t = {red = 'cherry', yellow = 'lemon', blue = 'grape'}
+
+        lu.assertTableContains(t, 'lemon')
+        lu.assertTableContains(t, 'grape')
+        lu.assertTableContains(t, 'cherry')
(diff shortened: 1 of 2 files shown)


test/test_luaunit.lua
```diff
@@ -1166,6 +1166,30 @@ TestLuaUnitAssertions = { __class__ = 'TestLuaUnitAssertions' }
...
+        local t = {3, 'some value', 1, 2}
+
+        assertFailure(lu.assertContains, t, 0)
+        lu.assertContains(t, 1)
+        lu.assertContains(t, 2)
+        lu.assertContains(t, 3)
+        assertFailure(lu.assertContains, t, 4)
+        lu.assertContains(t, 'some value')
+        assertFailure(lu.assertContains, t, 'other value')
+    end
```
bluebird75: I would like to see also some tests on a non-array table, just to make sure that we are not using ipairs instead of pairs in some places

test/test_luaunit.lua
```diff
@@ -1166,6 +1166,30 @@ TestLuaUnitAssertions = { __class__ = 'TestLuaUnitAssertions' }
...
+        local t = {3, 'some value', 1, 2}
+
+        lu.assertNotContains(t, 0)
+        assertFailure(lu.assertNotContains, t, 1)
+        assertFailure(lu.assertNotContains, t, 2)
+        assertFailure(lu.assertNotContains, t, 3)
+        lu.assertNotContains(t, 4)
+        assertFailure(lu.assertNotContains, t, 'some value')
+        lu.assertNotContains(t, 'other value')
+    end
```
bluebird75: Same here : I would like to see also some tests on a non-array table, just to make sure that we are not using ipairs instead of pairs in some places

bluebird75: Oups, just saw the next commit which does exactly this.

test/test_luaunit.lua
bluebird75: Oups, just saw the next commit which does exactly this.

n1tehawk:
>I find the name table_keyof() not explicit enough. How about *table_findkeyof()* or *table_whichkeyof()* ?

I can live with *table_findkeyof()*. Will rename the function accordingly.

>For the assert naming, we have already *assertStrContains* and *assertStrIContains* . Maybe *assertTableContains* would be more consistent, no ? What do you think ?

Agreed. *assertContains* is probably too general, *assertTableContains* makes sense.

I'll update the PR accordingly, and while at it will squash (= integrate) the commit that adds the tests for non-array tables.

Regards, NiteHawk

PR #88, opened 2017-01-24
Finalize on #68

n1tehawk:
Lack of feedback (and rebasing) have caused pull request #68 to be in a more or less abandoned state. Merging this would integrate the remaining `almostEquals()` change (i.e. the removal of an implicit `marginBoost`), and allow to close the PR.

Note: Due to rounding errors, this **might break existing test cases**. The internal tests have been adjusted accordingly, but we still might want to advise users that they may have to replace `assertAlmostEquals(a, b, margin)` with `assertAlmostEquals(b - a, margin)` in some cases. The latter is "safe" in the sense that it still applies the default `EPSILON` (see also https://github.com/bluebird75/luaunit/pull/68#issuecomment-258392921).

Regards, NiteHawk

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -74,10 +74,10 @@ Options:
   -n, --name NAME:        For junit only, mandatory name of xml file
   -c, --count NUM:        Execute all tests NUM times, e.g. to trig the JIT
   -p, --pattern PATTERN:  Execute all test names matching the Lua PATTERN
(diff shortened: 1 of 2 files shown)


bluebird75:
AppVeyor build fails just because it failed to download lua, which happens on a regular basis. We should have ways to detect this slightly better and abort at the right step.

PR #91, opened 2017-02-15
Use appveyor cache

bluebird75:
Reduce build time on AppVeyor by 60% by caching the lua for the different configurations. This also avoids build failing because of lua download failure (happening on a regular basis). Latest build on AppVeyor is now 1 min 21s


bluebird75:
@n1tehawk , I'll merge that one. Feel free to comment and review it afterwards.

PR #93, opened 2017-03-07
better referential transparency of tests

ldeniau:
The actual self in a test is a the _global_ table holding all the methods, and a very bad practice is to use it inside setup and teardown as it may provoke accidentally side effects between tests. I propose to create a new virgin instance of self for each test and use delegation (metatable with __index pointing to the class) to emulate the behaviour of the class.  No performance loss is measured, and the framework is much safer and sane. Best, Laurent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Documentation is available on
 no longer exported directly to the global namespace. See [documentation](http://luaunit.readthedocs.io/en/latest/#luaunit-global-asserts) on how to adjust or restore previous behavior.
 
 
-##Install
+## Install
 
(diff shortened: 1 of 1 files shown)


ldeniau:
It seems that I have conflicts (possibly coming form assertion counting), so the interesting code for this feature is inside the test loop in M.LuaUnit:execOneFunction:

            self.currentCount = iter_n

!            -- ensure referencial transparency of classes (local change)
!            if classInstance then
!                classInstance = setmetatable({}, {__index=classInstance})
!            end

            -- run setUp first (if any)
            if classInstance then

Issue #97, opened 2017-07-17
Add custom assertions


icamys:
It would be very nice to create custom assertions like this:

```
local luaunit = require('luaunit')

local msg = "Failed assertion on data set #%s"
local data = { ... }

for k, v in pairs(data) do
    -- the fourth parameter is custom assertion message
    luaunit.assertEquals(v.actual, v.expected, true, string.format(msg, k))
end
```

bluebird75:
I need to give it some thought. I find the default assertion messages quite good already, not sure we need to have custom messages.

But for assertions inside a loop, you do have a point.

icamys:
@bluebird75 The point is that if we have a fixture data set of 100 entries and we use a loop to run a test with those fixture entries and the test fails on 85th entry, it would be great to add to message at least the number of the entry.
At the moment we need to eliminate each of 100 entries to find the one that causes test failure

bluebird75:
I am working on my own patch for this, with small changes to what you provided. I want to review this deeply because it affects a big number of functions.

PR #98, opened 2017-07-17
Added custom assert messages

icamys:
Implemented feature mentioned in #97. Additional tests needed. I will add them if the PR will be accepted in future by repo owner

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -1111,6 +1111,13 @@ local function errorMsgEquality(actual, expected, doDeepAnalysis)
                          prettystr(expected), prettystr(actual))
 end
 
+local function createCustomMsg(descr, message)
+  if type(message) == 'string' then
+    return message .. "\n" .. descr
+  end
+  return descr
+end
+
 function M.assertError(f, ...)
     -- assert that calling f with the arguments will raise an error
     -- example: assertError( f, 1, 2 ) => f(1,2) should generate an error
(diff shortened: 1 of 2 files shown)


icamys:
Added all necessary tests. Please check the PR. I will be glad to listen to your comments.

bluebird75:
First of all, thanks for your effort on this, it is always appreciated when contributions are proposed.

I am going to accept the general contributions, after some adjustements : 

1. Please remove the removal of spaces at the end of lines. It is a small cosmetic improvement which makes your patch more tedious to read. If those spaces bother you, please submit another PR dedicated to cosmetic improvements.

2. I don't like the fact that the same code is repeated over and over to include the optional message. You can certainly craft a function to take care of this and have a smaller patch, with less copy/paste.

Apart from that, it's great that you also updated the tests, and espcially have added some missing tests.

Cheers,

Philippe

icamys:
@bluebird75 Hi, Phillip! Thanks for your fair remarks. I've updated my commit taking into account your requests.

juodumas:
Asserting in a loop (for example a parsing function) is impossible without this change. Please consider merging this for 3.3.

bluebird75:
I did a reimplementation around March 4. See 3d17658eff11a733d31e388e4f80265114212741 

This is now fully supported and documented and integrated into v3.3

PR #100, opened 2017-09-08
Better referential transparency of tests

ldeniau:
This is a very important request! Unfortunately, I renamed my dev branch in my fork that automatically closed the PR... Here is a "resurrection" of the PR referring to the right branch in my fork with some more comments.

The actual self in a test is a the global table holding all the methods, and a very bad practice is to use it inside setup and teardown as it may provoke accidentally side effects between tests. I propose to create a new virgin instance of self for each test and use delegation (metatable with __index pointing to the class) to emulate the behaviour of the class. No performance loss is measured, and the framework is much safer and sane.

The current PR now proposes a bit more than the point mentioned above, namely "assertion counting with pretty printing (right after test counts)", and a generalisation of assertAlmostEquals to tables "assertAllAlmostEquals" than make writing test on numerical content (e.g. vectors or matrices) much more compact. The differences with your master is very small, but in case of conflict, the important code to absolutely add to main master is the following (between tags) in method M.LuaUnit:execOneFunction, otherwise test may have easily side effects on each others!

        self.currentCount = iter_n

        -- PATCH
        -- ensure referencial transparency of classes (local change)
        if classInstance then
            classInstance = setmetatable({}, {__index=classInstance})
        end
        -- END OF PATCH

        -- run setUp first (if any)
        if classInstance then

 Best, Laurent.


ldeniau:
Bug fix about recursion in __index of classInstance metatable when combined with counts...

-- PATCH
    local node = self.result.currentNode
+  local class_mt = classInstance and {__index=classInstance}

[...]

+    -- ensure referencial transparency of classes (local change)
+   if class_mt then
+     classInstance = setmetatable({}, class_mt)
+   end

     -- run setUp first (if any)
     if classInstance then
-- END OF PATCH

bluebird75:
I won't go for the assert counting. I am still considering the other change, using delegation instead of the actual class. I don't see the scenario which can break the current situation and which is improved by this fix.

bluebird75:
I just added support for table in assertAlmostEquals() (starting from rev e859175533884ec836e04047c839d9f5733b372e ).

This obsoletes the (otherwise nice) proposition of this commit.

I am still not sold on full test isolation, so closing this PR now.

Issue #104, opened 2017-12-28
allow assertError* functions to operate correctly on non-string error messages


dmjacobsen:
Hello,

Since reviewing this (https://www.lua.org/wshop06/Belmonte.pdf) document, I've been using table (class instances, really) as my favorite "error message".  I do implement __tostring in these so it is fully possible to convert these messages to strings which luaunit can perform sensible comparisons against.

I'll submit a PR in the next few minutes to show you what I'm thinking of.

Thanks,
Doug

PR #105, opened 2017-12-28
allow the assertError* functions to accept non-string errors assuming…

dmjacobsen:
… that tostring can convert them (assuming that, for example, the __tostring function is set in an error table metatable)

This is related to issue #104.

### DIFF
--- a/test/test_luaunit.lua
+++ b/test/test_luaunit.lua
@@ -1948,7 +1948,7 @@ TestLuaUnitAssertionsError = {}
             id = 10,
         }) end)
 
-        lu.assertErrorMsgEquals("500", function() error(500) end)
+        lu.assertErrorMsgEquals("500", function() error(500, 2) end)
 
         assertFailure( lu.assertErrorMsgEquals, ' This is an error', self.f_with_error, x )
         assertFailure( lu.assertErrorMsgEquals, 'This .. an error', self.f_with_error, x )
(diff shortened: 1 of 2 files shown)


bluebird75:
Thank you for the contribution. It seems to break on Lua 5.1 and 5.2 according to travis ci. Could you have a look please ?

Thanks.

test/test_luaunit.lua
```diff
@@ -1947,6 +1948,8 @@ TestLuaUnitAssertionsError = {}
             id = 10,
         }) end)
 
+        lu.assertErrorMsgEquals("500", function() error(500) end)
+
```
bluebird75: try with error(500, 2). I expect that this would fix the problem for lua 5.1 and 5.2 . At least, this is the only difference that I could spot versus self.f_with_error()

test/test_luaunit.lua
```diff
@@ -1957,6 +1960,7 @@ TestLuaUnitAssertionsError = {}
         assertFailure( lu.assertErrorMsgMatches, 'is an err', self.f_with_error, x )
         lu.assertErrorMsgMatches( 'This is an error', self.f_with_error, x )
         lu.assertErrorMsgMatches( 'This is .. error', self.f_with_error, x )
+        lu.assertErrorMsgMatches("^500$", function() error(500) end)
```
bluebird75: try with error(500, 2). I expect that this would fix the problem for lua 5.1 and 5.2 . At least, this is the only difference that I could spot versus self.f_with_error()

dmjacobsen:
Cool, that did the trick.

bluebird75:
Merged.

Thanks for the contribution. Out of curiosity, could you tell me in what kind of project you use luaunit ? And do you feel about it ? Any things that I should add, remove or improve, besides assertError accepting table arguments ?

dmjacobsen:
Thank you!

Slurm (a high performance computing resource manager and scheduler, https://github.com/schedmd/slurm) provides a lua interface for implementing certain types of plugins.  To provide a rational engineering platform for these plugins I've stubbed out most of the interface that slurm provides and use luaunit to perform unit testing of all the functionality of my custom plugins.  This saves me a lot of time and effort by avoiding huge amounts integration testing involving a distributed system.  So far luaunit is certainly meeting my needs (I really almost exclusively use assertEquals and assertNil, but need to expand that further in time).  I'll be in touch if I have ideas for further improvements.

bluebird75:
Wow, I checked the web page, this quite a project ! I never thought that my little test library would end up in such high end computing product :-)

PR #109, opened 2018-06-04
Add assertErrorMsgContentEquals

okkez:
Both assertErrorMsgContains and assertErrorMsgMatches matches
the expected error message if the actual error message includes extra
things. The new function assertErrorMsgContentEquals matches the
expected error message only. It will fail assertion if actual error
message includes extra things.

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -1451,6 +1451,21 @@ function M.assertErrorMsgMatches( expectedMsg, func, ... )
     end
 end
 
+function M.assertErrorMsgContentEquals(expected_message, func, ...)
+    local success, error_message = pcall(func, ...)
+    if success then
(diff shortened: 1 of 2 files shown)


bluebird75:
Thanks for the proposal. Could you reorganise it to : 
1. change existing assertErrorMsgEquals() to a call to a new function _assertErrorMsgEquals( stripFileAndLine, expectedMsg, func, ...) with first arg set to false
2. write the new _assertErrorMsgEquals( stripFileAndLine, expectedMsg, func, ...) with the first argument being a boolean. When true, this will strip the file and line info as you do in your proposal
3. introduce the new function.

I like the idea. Maybe, I should even have designed assertErrorMsgEquals as you proposed...

okkez:
@bluebird75 I've created #110, could you check it?

PR #110, opened 2018-06-13
Reorganize adding assertErrorMsgContentEquals

okkez:
Both assertErrorMsgContains and assertErrorMsgMatches matches
the expected error message if the actual error message includes extra
things. The new function assertErrorMsgContentEquals matches the
expected error message only. It will fail assertion if actual error
message includes extra things.

See also #109


bluebird75:
thanks, looks great!

okkez:
Thank you for merging!

Issue #106, opened 2018-04-06
Allow calling `error()` with a non-string value.


fujimotos:
### Problem

Currently, luaunit assumes that `error()` function always takes a string
value as its first argument and performs some formattings based on the
assumption.

The problem is that we often pass non-string values to `error()` function
(e.g. passing details of the error using a table) and it breaks the execution
of luaunit.

### Example

Let's assume we have the following `error` call:

```lua
error({message = "xxx", code = 111})
```

This `error()` call will break the test execution with the following error:

```
$ luajit test_example.lua
luajit: /usr/local/share/lua/5.1/luaunit.lua:1911: attempt to call method 'gsub' (a nil value)
stack traceback:
    /usr/local/share/lua/5.1/luaunit.lua:1911: in function 'protectedCall'
    /usr/local/share/lua/5.1/luaunit.lua:1967: in function 'execOneFunction'
    /usr/local/share/lua/5.1/luaunit.lua:2062: in function 'runSuiteByInstances'
    /usr/local/share/lua/5.1/luaunit.lua:2122: in function 'runSuiteByNames'
    /usr/local/share/lua/5.1/luaunit.lua:2185: in function 'run'
    test/run-test.lua:12: in main chunk
    [C]: at 0x55942f0111d0
```

### Desired behaviour

No error should be raised on non-string error values.

bluebird75:
Indeed, we fixed the assertError() functions for this but I did not thing about general error catching. I'll work on a fix and let you know. Or feel free to propose one :-)   .

PR #113, opened 2018-06-21
Add support for all error() argument types

javabird25:
Fix bluebird75/luaunit#106.
luaunit crashes when `error()` called with a non-string argument, I fixed this.

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -504,6 +504,12 @@ function M.adjust_err_msg_with_iter( err_msg, iter_msg )
 
     local RE_FILE_LINE = '.*:%d+: '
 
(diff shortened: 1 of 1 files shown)


bluebird75:
Thanks for the proposal. Please also add some test cases for this.

bluebird75:
Sorry for the long time without response. Your PR contains a lot of changes related to whitespace at the end of the lines (probably your editor settings). While this could be an interesting style fix, I don't want to have it in this PR as it is creating noise which makes finding the real changes difficult.

Regarding your test, could you add a verification of the error message content ?

Thanks in advance.

PR #117, opened 2018-09-27
Flush output in the default verbosity

kou:
In the default verbosity, LuaUnit outputs one character such as "."
and "E". We can't know the one character progress until all tests are
finished or the stdout internal buffer is fulled because the one
characters are buffered in the stdout internal buffer.

We can the one character progress immediately by flushing the stdout
internal buffer.

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -2190,6 +2190,7 @@ TextOutput.__class__ = 'TextOutput'
                 io.stdout:write("Ok\n")
             else
                 io.stdout:write(".")
+                io.stdout:flush()
             end
         else
             if self.verbosity > M.VERBOSITY_DEFAULT then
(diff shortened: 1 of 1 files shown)


kou:
Ah, sorry. I missed describing my environment.
I'm using Debian GNU/Linux.

bluebird75:
Thanks for the contribution.

Out of curiosity, in what project are you using luaunit ?

kou:
Thanks.

We're using LuaUnit in the following projects:

  * [XMLua](https://github.com/clear-code/xmlua) 
  * [LuaCS](https://github.com/clear-code/luacs)
  * [LuaWebDriver](https://gitlab.com/clear-code/lua-web-driver)

Issue #118, opened 2019-03-14
assertIs()/assertNotIs() may cause error when failing in comparing tables with protected metatables


seahorse47:
I use `assertIs()`/`assertNotIs()` to verify if two object references (actually two tables with protected metatables) are the same / different. When it fails, I get error `cannot change a protected metatable` in `_table_raw_tostring()` function. This is caused by `prettystrPairs()` function calls when `assertIs()`/`assertNotIs()` fails. Here is the example code:

```
local lu = require "luaunit"

-- To make it easier to reproduce this error, enable `PRINT_TABLE_REF_IN_ERROR_MSG`
lu.PRINT_TABLE_REF_IN_ERROR_MSG = true

local Object = {}
function Object:new(...)
    return setmetatable({
        class = self,
    }, {
        __metatable = "private",
        __index = self,
    })
end

function testA()
    local obj1 = Object:new()
    local obj2 = Object:new()
    lu.assertIs(obj1, obj2) -- will fail
end

local runner = lu.LuaUnit.new()
runner:setOutputType("text")
os.exit(runner:runSuite())
```

P.S.
When I use `assertIs()` on tables, I don't really care about the contents in the tables. I do only care whether they are the same. `assertEquals()` may be a better choice when I care about the contents.**Is it possible to show only table references in fail messages of `assertIs()`/`assertNotIs()`?**

bluebird75:
Thanks for the report and the Pull Request. I am quite slow those days to react, sorry for that. I need more time to understand the proposal because it touches parts of the code which I have not written myself.

I'll let you know of the progress.

bluebird75:
Ok, I get it. The misbehaving code was trying to get the reference value of the table, when the metatable has a __tostring(). Now the question is: how do I obtain the reference of a table, whose metatable is both protected and contains a __tostring() converter ?

I don't see how to do that right now, but in the meantime, I can still fix your reported bug.

Thanks a lot.

bluebird75:
fixed in 2d85352c92303a4032ff5fbde782cb9e765c3f9d

PR #119, opened 2019-03-14
Fix `_table_raw_tostring()` error and show brief string for values in fail messages of `assertIs()`/`assertNotIs()`.

seahorse47:
Related issue: #118


bluebird75:
I did it with a very slightly different approach, but I would not have found it without your contribution.

I forced printing of table ref for assertIs() and assertNotIs(), it makes much more sense indeed

PR #123, opened 2019-05-13
Add GLOBAL_TESTS option

printercu:
To be able to define test outside of global namespace

Hi!

With this patch it's possible to define tests and groups outside of global namespace.

```lua
local lu = require('luaunit')
local t = lu.group('SomeFeauture')

t.setup = function()....
t.testSmth = function() ...
```

It's a good practice to not introduce unnecessary global variables and luacheck warns on this by default.

### DIFF
--- a/TODO.txt
+++ b/TODO.txt
@@ -65,7 +65,7 @@ Version 3.4:
 - catch the os.exit call with a handler or something, to avoid aborting the tests
 - move all file:line description to stack trace
 - better deal with one line formatting
-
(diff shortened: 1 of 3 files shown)


bluebird75:
I was planning to be able to define tests to run as a list at some point. I am not sure that I want to replace a full global namespace with a module global. I would much rather have a list of tests which we can pass around and use.

Do you think you could work on another prototype along the lines of what I propose ?

printercu:
Can you please provide a sample use case?

PR #130, opened 2020-07-12
Adapt randomseed to Lua 5.4

javalikescript:
The `math.randomseed` method has been updated in _Lua_ 5.4, see [Lua reference](https://www.lua.org/manual/5.4/manual.html#pdf-math.randomseed).
The current _luaunit_ usage fails occasionally with the error "_bad argument # 1 to 'randomseed' (number has no integer representation)_".
This pull request will call the method without arguments when using _Lua_ 5.4.

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -213,11 +213,7 @@ end
 M.private.sortedPairs = sortedPairs
 
 -- seed the random with a strongly varying seed
-if _VERSION <= "Lua 5.3" then
-    math.randomseed(os.clock()*1E11)
-else
-    math.randomseed()
-end
+math.randomseed(math.floor(os.clock()*1E11))
 
 local function randomizeTable( t )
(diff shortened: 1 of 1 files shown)


javalikescript:
For the sake of the coverage, I modified the pull request to avoid testing the Lua version.
The bad part is that it not more use the new Lua 5.4 random seed.

bluebird75:
The decrease in coverage is just because I don't have lua 5.4 in the CI pipeline yet. Don't worry about that.

Ordering test randomly does not need a strong random generator though, I believe we can live with suboptimal seed :-)

javalikescript:
I am in favor of the new Lua 5.4 random seed but I do not like the Lua version test.
The seed using a single line also helps to read the code so I will stick to it, for now.

bluebird75:
Thanks. I'll try to add proper support for Lua 5.4

PR #132, opened 2020-10-29
travis: fix [ usage

kevans91:
The || operator is actually a shell construct, not recognized by test/[. We
could change this to [ expr ] || [ expr2 ], but it's cheaper to just use
the POSIX-specified -o instead.

Signed-off-by: Kyle Evans <kevans@FreeBSD.org>

--

Noticed while looking at the build log, in before_install.1 of the LUANUMBER=float jobs: 

+'[' lua5.2 == lua5.3
.travis/setup_lua.sh: line 110: [: missing `]'
+lua5.2 == lua5.4 ']'
.travis/setup_lua.sh: line 110: lua5.2: command not found

### DIFF
--- a/.travis/setup_lua.sh
+++ b/.travis/setup_lua.sh
@@ -107,7 +107,7 @@ else # -e $LUA_HOME_DIR
 
         # adjust numerical precision if requested with LUANUMBER=float
         if [ "$LUANUMBER" == "float" ]; then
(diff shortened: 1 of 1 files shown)


bluebird75:
Thanks. I always run into trouble with shell condition testing. This is really arcane ! In 20 years, I still have not figured it out

PR #133, opened 2020-10-29
Allow the output format to be specified by the environment

kevans91:
Described in a little more in-depth in the commit messages, but this small PR would be quite appreciated -- this will check for LUAUNIT_DEFAULT_OUTPUT in the environment and use that as if it were supplied as the --output option in the absence of an overriding --output option from the arguments. This opens us up to using a shebang line to fully specify how the test is generally desired to be read:

```
#!/usr/bin/env -S LUAUNIT_DEFAULT_OUTPUT=TAP /usr/local/bin/lua53
```

Specifically, I'd like to import luaunit into FreeBSD's base system for unit testing of the lua bindings we're writing for various base libs, and our infrastructure can/will insert effectively the above shebang (except using our base lua interpreter) so that the test driver (Kyua) can just execute this test like it does any other test and examine the output without having to understand that it's a lua script or anything else.

A functional smoke test has been added to insert LUAUNIT_DEFAULT_OUTPUT into the environment for one of the TAP tests that's already passed to try and make it clear if any failure is due to a regression in the TAP output vs. the new environment variable. TAP was randomly chosen for this example because it was the first non-default format test that I came across.

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -3239,6 +3239,8 @@ end
         self.patternIncludeFilter = options.pattern
         self.shuffle              = options.shuffle
 
+        options.output     = options.output or os.getenv('LUAUNIT_DEFAULT_OUTPUT')
+
         if options.output then
             if options.output:lower() == 'junit' and options.fname == nil then
                 print('With junit output, a filename must be supplied with -n or --name')
--- a/run_functional_tests.lua
+++ b/run_functional_tests.lua
@@ -182,7 +182,11 @@ local function check_tap_output( fileToRun, options, output, refOutput, refExitC
     if not envOptions:match("LUAUNIT_DEFAULT_OUTPUT[^%w]") then
         outputArg = '--output TAP'
     end
-    osExpectedCodeExec(refExitCode, '/usr/bin/env -S %s %s %s %s %s > %s',
+    if envOptions ~= '' then
+        envOptions = '/usr/bin/env -S ' .. envOptions
+    end
+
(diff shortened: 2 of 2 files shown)


kevans91:
It looks like I'll need to reimagine the test change, since I forgot the test needs to be able to run on Windows as well...

bluebird75:
Nice. Can you tell me a little bit more about what you are testing with luaunit, out of curiosity ?

kevans91:
> Nice. Can you tell me a little bit more about what you are testing with luaunit, out of curiosity ?

Sure! We imported Lua into the base distribution of FreeBSD, and now we're writing bindings for various system libs so we can write some utilities in Lua. e.g. we produced some bindings for libjail that does operations on jails (much like containers, if you're not familiar), and we're needing tests for the various bindings we're producing before we start writing system utilities that depend on them.

bluebird75:
I am interested in more feedback on using lua with many tests. I believe the current setup of declaring global tests is osolete and also luaunit should provide more convenience on running a test module with one or several test cases.

Don't hesitate to share your ideas or paint point in the process !

kevans91:
Sure, will do! I tend to agree- our C/C++/sh testing frameworks all typically require you to specify the test to run and also drive cleanup independently -- I tend to not care about driving cleanup independently, but it is nice to be able to run a single test case in isolation.

The downside is their output formats are really bad. I really like luaunit's default output format and how it actually provides useful information about failures.

bluebird75:
Note that I changed the variable to LUAUNIT_OUTPUT for more simplicity.

PR #134, opened 2020-11-21
use black to pass flake8 for py file

waketzheng:
Use `black` to  reformat python file
```
pip install --user toml black flake8
black doit.py
```

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,0 @@
-[flake8]
-max-line-length = 150
-extend-ignore = E401
-exclude = doc/
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,4 @@
+[flake8]
+max-line-length = 150
+extend-ignore = E401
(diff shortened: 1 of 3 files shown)


bluebird75:
Thanks for the contribution.

Can you remove the pyproject file ? I don't see the interest here

waketzheng:
Nice. I use `.flake8` to replace pyproject file, and `flake8` complaint nothing now.

bluebird75:
I don't want to introduce new files. There are already too many files at the root of this project. So, can we have it without the .flake8 file ?

waketzheng:
Sure~
And you can run the following command at the root of this project. So that `flake8` will pass next time.
```bash
echo -e '[flake8]\nmax-line-length=150\nextend-ignore=E401\nexclude=doc/' > .flake8 && echo .flake8 >> .git/info/exclude
```

PR #135, opened 2020-11-25
Lua 5.4 compatibility

mna:
Hello,

As far as I can tell from testing it on one of my projects, luaunit works fine on Lua 5.4. This change would make it possible to install with that version (if merged, this change should be published on luarocks, of course). Of course, LuaUnit's tests themselves pass fine with Lua 5.4.1 that I have on my system (Linux Fedora).

Thanks,
Martin

### DIFF
--- a/luaunit-3.3-1.rockspec
+++ b/luaunit-3.3-1.rockspec
@@ -34,7 +34,7 @@ description =
 
 dependencies =
 {
-	"lua >= 5.1", "lua < 5.4"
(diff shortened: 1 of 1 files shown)


bluebird75:
I have setup successfully lua 5.4 on travis-ci successfully. So yes, this is good.

I just need to actually do another release to have this officially propagated.

mna:
Génial, merci!

PR #155, opened 2023-01-02
Add method and test prefix options.

obszczymucha:
This allows specifying the prefix for test and method names, so the tests can be now defined as follows:

```
function shouldValidateArguments()
...
end
```

or

```
function should_validate_arguments()
...
end
```

If not provided, the behavior is unchanged and it defaults to the current `test` prefix.

Example:
```
local runner = lu.LuaUnit.new()
runner:setOutputType( "text" )
os.exit( runner:runSuite( "-t", "should" ) )
```


luaunit.lua
```diff
@@ -3393,6 +3424,9 @@ end
         self.exeRepeat            = options.exeRepeat
         self.patternIncludeFilter = options.pattern
         self.shuffle              = options.shuffle
+        self.methodPrefix         = self.methodPrefix or options.methodPrefix
+        self.testPrefix           = self.testPrefix or options.testPrefix
+        self.testSuffix           = self.testSuffix or options.testSuffix
```
bluebird75, line 3430: Not sure about that part. The idea normally is that options passed on the command-line should overwrite what you defined in your test runner. This gives a dynamic behavior on the command-line, and I have the feeling that this code does not respect this behavior.

bluebird75:
Hi,

Sorry for the delay, I just moved to a new house so things like open source have been left behind.

I like the patch, I need to further review it but it looks great. And with unit-tests, this is exceptional 👍

obszczymucha:
Hey. No worries at all! Thanks for responding!

luaunit.lua
obszczymucha: Would you be able to give an example on how can I set up the options in the test runner?
When I tried doing this:

```
local runner = lu.LuaUnit.new()
runner:setOutputType( "text" )
os.exit( runner:runSuite( "-p", "somepattern" ) )
```

and then override it via the command line:

```
lua my_test.lua -p somenewpattern
```

It doesn't override it. It still uses the `somepattern`.

bluebird75: Yeah, I reviewed the code and indeed, it is imperfect. Setting up the arguments directly in function completely erase the command-line argument parsing.

I am not sure what's the best way to fix this general problem. My idea is that command-line should override function parameters when an option is defined in both places, but for non conflicting options, both should be merged.

PR #174, opened 2026-06-11
respect metatable's __eq in luaunit assertEquals

szabolotskikh-inmation:
If there is `__eq` in a table's metatable, then `luaunit.assertEquals` does not take it into account.
However, in real projects it often happens that object comparison shall be done in accordance with special rules, and one cannot simply expose all key-values of a table because of several reasons, e.g., they break JSON encoding or there is some special userdata, etc.

According to the documentation https://www.lua.org/manual/5.5/manual.html#2.4
> **__eq:** the equal (==) operation. Behavior similar to the addition operation, except that Lua will try a metamethod only when the values being compared are either both tables or both full userdata and they are not primitively equal. The result of the call is always converted to a boolean.
**__add:** the addition (`+`) operation. If any operand for an addition is not a number, Lua will try to call a metamethod. It starts by checking the first operand (even if it is a number); if that operand does not define a metamethod for `__add`, then Lua will check the second operand. If Lua can find a metamethod, it calls the metamethod with the two operands as arguments, and the result of the call (adjusted to one value) is the result of the operation. Otherwise, if no metamethod is found, Lua raises an error.

I am proposing that if there is `getmetatable(t).__eq` either for `actual` or `expected` Lua tables then luaunit doesn't manually compare key-value pairs. If it somehow breaks somebody's code then there is the Boolean option `RESPECT_METATABLE_EQUALS` (default `true`) to disable such behavior.

### DIFF
--- a/luaunit.lua
+++ b/luaunit.lua
@@ -29,7 +29,6 @@ M.PRINT_TABLE_REF_IN_ERROR_MSG = false
 M.LINE_LENGTH = 80
 M.TABLE_DIFF_ANALYSIS_THRESHOLD = 10    -- display deep analysis for more than 10 items
 M.LIST_DIFF_ANALYSIS_THRESHOLD  = 10    -- display deep analysis for more than 10 items
-M.RESPECT_METATABLE_EQUALS = true
 
 -- this setting allow to remove entries from the stack-trace, for 
 -- example to hide a call to a framework which would be calling luaunit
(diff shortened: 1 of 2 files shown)


luaunit.lua
```diff
@@ -1259,6 +1260,22 @@ local function _is_table_equals(actual, expected, cycleDetectTable, marginForAlm
         return actual == expected
     end
 
+    if M.RESPECT_METATABLE_EQUALS then
+        if actual == expected then
```
bluebird75: The code here does not make sense to me.

A sensible approach would be: if one of the table has a metatable with __eq, let's compare the tables directly and let lua handle the comparison properly.

If not metatable present or no __eq present, let's stick to regular table deep comparison.

And no need for an extra flag, the behavior makes sense in all situation

szabolotskikh-inmation: > The code here does not make sense to me.

I thought that it could at least partially support the case when `__metatable` is set too. I have added tests indicating the weird behavior.

I have removed that comparison, now it's more consistent.

